PR #8, opened 2022-06-01
remove vscode settings

harish-sethuraman:
Don't add vs code settings my fren


bigint:
Thanks for the PR, this is merged in GitLab 💟

PR #23, opened 2022-07-20
chore: add lint and prettier workflow

harish-sethuraman:
Added workflow for lint and prettier Should we add separate ones for tests?


bigint:
@harish-sethuraman can we use yarn 🙏🏻

PR #78, opened 2022-07-21
chore: remove vs code setttings

harish-sethuraman:


bigint:
Sed :( can we leave extensions?

harish-sethuraman:
Umm Okay. But I don't think that'd be useful anyways

bigint:
Makes sense 👍🏻

PR #153, opened 2022-07-28
chore: add `no-use-before-define`

harish-sethuraman:
Added `no-use-before-define` rule and fixed breakage


bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #198, opened 2022-07-31
fix: retain logged in session

0xkeivin:
## Current Behavior
Whenever a user loads a new tab on lenster, it will trigger the `logout()` function due to a change in `connector` 

## Expected Behavior
Users should be able to retain their logged in sessions even if new tabs are opened. 

## Steps to reproduce
Load lenster on a new tab and it will logout the user 

## What platform(s) does this occur on?
Web - Chrome, Firefox 

## Notes
- Wanted to merge on testnet but seems it is lagging several commits behind


bigint:
@0xkeivin We are listening to connector because when user switches between accounts or network we need to logout them 🙇🏼 

Do u have any other ideas to fix it 🤔 

https://user-images.githubusercontent.com/69431456/182032516-c23c352c-6b0c-484a-8325-20d4b672a32b.mov

bigint:
@0xkeivin still seems like user is not logging out 😕, is this PR still WIP?

0xkeivin:
have converted to draft. 🙏🏻  
let me test further on this.

0xkeivin:
Have tested the following: 

begin with a valid login session:
1. Changing chain network
- causes `chain?.id` does not meet the original set of IF conditions and logs out the user
2. Changing to different polygon wallet 
- causes `storedAddress` to be undefined and logs out
3. Opening new tab/windows in same browser
- `storedAddress` matches current `address` hence it won't log out the user
4. refreshing the page 
- user still logged in 
5. locking metamask
- user will be logged out

Notes:
- unable to make it work with `connector` within useEffect()  as several actions will cause it to change (i.e new tab/window) 😓

src/components/SiteLayout.tsx
```diff
@@ -82,6 +82,10 @@ const SiteLayout: FC<Props> = ({ children }) => {
   useEffect(() => {
     const accessToken = Cookies.get('accessToken')
     const refreshToken = Cookies.get('refreshToken')
+    const storedUserState = localStorage.getItem('lenster.store' || '{}')
```
bigint, line 85: **question:** @0xkeivin we can get this value from the store `usePersistStore` right? any reason we are using localstorage?

0xkeivin: have overlooked that. 😓 
let me try replacing it with values from `useAppPersistStore`

0xkeivin: have updated in latest commit

bigint (approved):
**LGTM** :shipit: 

Rad LGTM

bigint:
Hey fren, you are now eligible for Gitpoap for contributing to Lenster 🌸, go and claim your beautiful POAP 🏃🏼‍♂️ 

https://www.gitpoap.io/gh/lensterxyz/lenster

Issue #66, opened 2022-07-20
Show opensea address once the collection is done


bigint:

kulot1327:
hello

bigint:
Not possible for now 😢

PR #344, opened 2022-08-09
feat: hide communities from top collected

### LINKED ISSUE #36: [Enhancement] Hide communities from top collected

andriishupta:
CodeSandbox logoCodeSandbox logo&nbsp; Open in CodeSandbox <a href="https://codesandbox.io/p/github/andriishupta/lenster/feat/36-Hide_communities_from_top_collected?workspace=%7B%22gitSidebarPanel%22:%22PR%22,%22sidebarPanel%22:%22GIT%22%7D">Web Editor</a> | <a href="https://codesandbox.io/p/vscode?owner=andriishupta&repo=lenster&branch=feat/36-Hide_communities_from_top_collected">VS Code</a>

- add query sources for TOP_COLLECTED
- add `lensterCommunity` app to keep reference that it exists
- add .gitignore for .idea
- fix startup issue with mixpanel.ts

Closes #36

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ tsconfig.tsbuildinfo
 
 # Sentry
 .sentryclirc
+
+# IDE
+.idea
--- a/data/apps.ts
+++ b/data/apps.ts
@@ -35,3 +35,9 @@ export const apps = [
     logo: 'lenstube.svg'
   }
 ];
+
+export const lensterCommunity = {
+  id: 'lenster community',
+  name: 'Lenster Community',
+  logo: 'lenster.svg'
+};
--- a/src/components/Explore/Feed.tsx
+++ b/src/components/Explore/Feed.tsx
@@ -12,6 +12,7 @@ import { MirrorFields } from '@gql/MirrorFields';
 import { PostFields } from '@gql/PostFields';
 import { CollectionIcon } from '@heroicons/react/outline';
 import { Mixpanel } from '@lib/mixpanel';
+import { apps } from 'data/apps';
 import React, { FC, useState } from 'react';
 import { useInView } from 'react-cool-inview';
 import { useAppPersistStore } from 'src/store/app';
@@ -51,12 +52,14 @@ interface Props {
 }
 
 const Feed: FC<Props> = ({ feedType = 'TOP_COMMENTED' }) => {
+  const sources = feedType === 'TOP_COLLECTED' ? apps.map((app) => app.id) : undefined;
   const currentUser = useAppPersistStore((state) => state.currentUser);
   const [publications, setPublications] = useState<LensterPublication[]>([]);
   const [pageInfo, setPageInfo] = useState<PaginatedResultInfo>();
   const { data, loading, error, fetchMore } = useQuery(EXPLORE_FEED_QUERY, {
     variables: {
       request: {
+        sources,
         sortCriteria: feedType,
         limit: 10,
         noRandomize: feedType === 'LATEST'
@@ -75,6 +78,7 @@ const Feed: FC<Props> = ({ feedType = 'TOP_COMMENTED' }) => {
       const { data } = await fetchMore({
         variables: {
           request: {
+            sources,
             sortCriteria: feedType,
             cursor: pageInfo?.next,
             limit: 10,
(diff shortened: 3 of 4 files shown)


andriishupta:
followed API to get every existing app that Lenster supports, except for "lenster community"
https://docs.lens.xyz/docs/explore-publications#request

data/apps.ts
```diff
@@ -35,3 +35,9 @@ export const apps = [
     logo: 'lenstube.svg'
   }
 ];
+
+export const lensterCommunity = {
+  id: 'lenster community',
+  name: 'Lenster Community',
+  logo: 'lenster.svg'
+};
```
bigint, line 43: **suggestion:** can we move this to `apps` object?

.gitignore
```diff
@@ -28,3 +28,6 @@ tsconfig.tsbuildinfo
 
 # Sentry
 .sentryclirc
+
+# IDE
+.idea
```
bigint, line 33:
```suggestion

```

bigint (changes requested):

.gitignore
andriishupta: I think you don't want me to commit my 9 files in .idea folders so that I would remove it :)

just to mention - a lot of scaffold projects include this ignore by default

bigint: Ahh gotcha, then let's go with yours 🚀

data/apps.ts
andriishupta: yes, we can:

- so, in Feed, I would filter it out
- and posts by the "lenster community" would look like this:

before vs. after
image

--

image

andriishupta: p.s. I specifically didn't want to add it to apps at first, cause checked that this will cause this change

also, is the same icon + name "Lenster Community" ok?

we could do the name "Lenster", so it would show same as lenster

bigint: But I'm not seeing `lensterCommunity` used anywhere 🤔 

Even if we add to `apps` it will show `Posted via Lenster Community` but we don't want to show it and it doesn't makes sense, it's just a community creation 🤔

bigint: Also we have lot of sources not listed in the file, it might be from small apps but still we want to show everything except the community creations

andriishupta: I left a message in the commit message and in the description that this lensterCommunity is only added for visibility that it exists - yes, I didn't use it :)

--

that was my other comment - "filter out" from apps

```
apps.map(app => app.id).filter(id => id !== 'lenster community')
```

--

so the next option would be to filter on the front-end and in bad cases:
- we query with a limit of 10 items
image
- we get 9 communities
- we show only 1 post in top collected

I haven't found any other way in API to filter, for example, by attributes

bigint: Yup we are blocked by the backend, you can see more here

https://github.com/aave/lens-api-examples/issues/48

andriishupta: that totally would be my solution :)

so, wait for now?
I can clean PR and leave it in a pending state

andriishupta: so, if Explore for Top Collected needs to have small "unregistered" apps to be present - I think we should wait 😬

I am totally against filtering on FE with the case that I described - you didn't propose, I didn't propose, but to be clear about it 😅

andriishupta:
@bigint, I have seen that you merged the main branch - I prefer to rebase and force-push in feature branches so that I will squash my commits and will force push new single commit when there would be new changes

why would I or some other contributor pollute git history with 10+ commits like "fix lint", "fix typo", etc.

bigint:
Yup I did rebase in GitHub UI 🙇🏼 

We always do squash merge PRs 🤗 

Also I'm going ahead and closing this PR because we don't want to filter with the apps only we specified 🙏🏼 

As said this is blocked on backend, when we get unblocked we can visit this issue back 🙏🏼

andriishupta:
@bigint cool ;)

good that I have checked the codebase and now have an overall understanding of how it is built :)

I checked also other issues today - for some reason, Post collection didn't work on Mainnet, so I cannot check issue #66  - "transaction was indexing" for 15min and I just closed the modal.
Others don't have descriptions yet, and I expect them to be bigger, so I could take them when you would add some details


And again, if you know the "implementation plan" or you know that it is blocked by BE - please, let me know

It was fun to do this PR today 😅

Thanks!

bigint:
Again #66 is blocked by Lens backend 😭 

I'd suggest you to DM me on Discord 💌 https://lenster.xyz/discord

PR #359, opened 2022-08-10
fix: infura f*er

bigint:


bigint:
Hope I didn't committed private key 🙈

joshbenaron:
I would change the Bundlr URL to https://node2.bundlr.network. If the tx (data payload) is <100KiB then the transaction is free on node2 😏

bigint:
@joshbenaron Done thanks fren 🚀 

https://github.com/lensterxyz/lenster/commit/bac9b0b3f771aa5fd074fbceb8aafe3f9a0f522c

PR #376, opened 2022-08-12
feat: add alert prompt before deleting the account

### LINKED ISSUE #63: Add alert prompt before deleting the account

andriishupta:
Fixes #63 

## Screenshots

image

image

### DIFF
--- a/src/components/Settings/Delete/index.tsx
+++ b/src/components/Settings/Delete/index.tsx
@@ -54,7 +54,7 @@ const CREATE_BURN_PROFILE_TYPED_DATA_MUTATION = gql`
 `;
 
 const DeleteSettings: FC = () => {
-  const [showWarningModal, setShowWarningModal] = useState(false);
+  const [showWarningModal, setShowWarningModal] = useState<boolean>(false);
   const userSigNonce = useAppStore((state) => state.userSigNonce);
   const setUserSigNonce = useAppStore((state) => state.setUserSigNonce);
   const setIsConnected = useAppPersistStore((state) => state.setIsConnected);
--- a/src/components/Settings/Delete/index.tsx
+++ b/src/components/Settings/Delete/index.tsx
@@ -4,15 +4,18 @@ import { GridItemEight, GridItemFour, GridLayout } from '@components/GridLayout'
 import UserProfile from '@components/Shared/UserProfile';
 import { Button } from '@components/UI/Button';
 import { Card, CardBody } from '@components/UI/Card';
+import { Modal } from '@components/UI/Modal';
 import { Spinner } from '@components/UI/Spinner';
+import { WarningMessage } from '@components/UI/WarningMessage';
 import Seo from '@components/utils/Seo';
 import { CreateBurnProfileBroadcastItemResult } from '@generated/types';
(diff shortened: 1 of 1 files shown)


andriishupta:
@bigint put PR in "Ready for review" such seems that you rebased it and deployed it and maybe wanted merge(not sure if you can update "Draft" status)

Please, let me know  if any updates to texts, logic are needed

thanks!

src/components/Settings/Delete/index.tsx
```diff
@@ -51,6 +54,7 @@ const CREATE_BURN_PROFILE_TYPED_DATA_MUTATION = gql`
 `;
 
 const DeleteSettings: FC = () => {
+  const [showWarningModal, setShowWarningModal] = useState(false);
```
bigint, line 57:
```suggestion
  const [showWarningModal, setShowWarningModal] = useState<boolean>(false);
```

src/components/Settings/Delete/index.tsx
```diff
@@ -133,6 +137,8 @@ const DeleteSettings: FC = () => {
     });
   };
 
+  const isDeleting = typedDataLoading || signLoading || writeLoading;
```
bigint, line 140: Kudos 👏🏼

bigint (approved):
**LGTM** :shipit: 

Rad LGTM

src/components/Settings/Delete/index.tsx
andriishupta: Thanks for comment!
I will count for future requests your preferred type annotation approach!


In some projects and by inner feeling it is better to avoid explicit typings if there is already 100% implicit type detected

TS/eslint even has error: Avoid explicit typing for implicitly derived
Something like this**

That is why I already prefer false, instead of <boolean>(null)

For simple types - it is easy to do
But if your function has complex Pick, Omit, Params, etc. - we usually believe IDE and don't annotate))

And don't worry - I use explicit typings for things like lodash, or Json transformations if "they"(IDE and TS itself) cannot derive type

🔥🤓

andriishupta:
Thanks for review! 🙇

andriishupta:
@bigint, and sometimes I will have the urge to justify/explain "why I did that" cause programming is like 50 shades of gray - everyone has own kink 🧨 depending on previous experience

But I am learning your preferred way of things 😉 

Thanks for merging!

src/components/Settings/Delete/index.tsx
andriishupta: @bigint found, it was "inferable", not "derived"

prefer "no inferrable types" rule - 🔥 

https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-inferrable-types.md

bigint: Oh this makes sense @andriishupta, we can make a separate PR to fix this 🚀

andriishupta: no need for separate PR - there are thousands of rules 🥵
just if you see sense it in - I will omit such things in "obvious" inferred places in my next PRs

thanks for checking this out 🤓

PR #575, opened 2022-08-30
Add manifest.json to support PWA mode for Lenster

royalaid:
## What does this PR do?

Adds a `manifest.json` file, allowing Lenster to be added to home screen for a more app-like UI, this could serve as a bridge while a real mobile app is built. Adding a service worker as a future enhancement would be the next step for a fully fledged PWA. I took some guesses with the icons which need to checked.

## Type of change

## How should this be tested?

- [ ] Run the app and attempt to add it to your home screen via chrome mobile

## Checklist

- I haven't commented my code, particularly in hard-to-understand areas
- I haven't checked if new and existing unit tests pass locally with my changes

### DIFF
--- a/public/manifest.json
+++ b/public/manifest.json
@@ -1,0 +1,26 @@
+{
+  "short_name": "Lenster",
+  "name": "Lenster.xyz",
+  "iconPath": "./favicon.svg",
+  "description": "Lenster is a decentralized, and permissionless social media app built with Lens Protocol",
+  "orientation": "portrait",
+  "display": "standalone",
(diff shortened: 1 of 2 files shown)


bigint:
Thanks for the contribution @royalaid 🙏🏼 

PWA mode will not be supported until serviceworkers are in place 🙇🏼 

Lenster was PWA long back but we removed it because our server bandwidth consumption was very high, we may consider this in future 🙏🏼

royalaid:
Sounds good, LMK if you want help in the future!

PR #731, opened 2022-10-04
feat: setup xmtp

bhavya2611:
This PR is setting up XMTP store, to have all data of messages and conversations in one place, this further is also setting up the route and will look into the streaming of messages and conversations.


bigint:
@bhavya2611 can you rebase the branch with main 🙏🏼

bigint:
Can we make use of `/messages` route which we already defined 🙇🏼

bigint:
I also invited you to this GitHub org so you can create branched directly from here 🚀

PR #734, opened 2022-10-04
chore: add lint github action

elisealix22:
## What does this PR do?

Adds a Github action that lints, typechecks, and checks formatting. We want to ensure we keep the repo clean if we can.


bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #779, opened 2022-10-10
feat: add message preview component

elisealix22:
## What does this PR do?
This PR maps Lens profiles to XMTP accounts and only shows messages for conversations that include a valid Lens profile.

This is a work in progress and behind the `messages` feature flag. Some known issues include:
- Right now the preview shows the first message not the most recent
- The conversations in the left panel are not sorted correctly
- Pieces are duplicated between Message and Messages. They should be pulled out into their own `ConversationList` component.

Screen Shot 2022-10-10 at 2 57 29 PM

## How should this be tested?

- Ensure you're in the FF and navigate to `Messages`
- You should see a list of all your current DM conversations with Lens profiles
- Tapping the cell should open the message list component


src/components/Messages/index.tsx
```diff
@@ -1,26 +1,66 @@
...
+  const setClient = useMessageStore((state) => state.setClient);
+  const isMessagesEnabled = isFeatureEnabled('messages', currentProfile?.id);
+
+  const conversations = useMessageStore((state) => state.conversations);
+  const setConversations = useMessageStore((state) => state.setConversations);
+  const messagePreviews = useMessageStore((state) => state.messagePreviews);
+  const setMessagePreviews = useMessageStore((state) => state.setMessagePreviews);
+
+  const peerAddresses = Array.from(messagePreviews.keys());
+  const { error: profilesError } = useQuery(ProfilesDocument, {
```
bhavya2611, line 43: We should pull this out in a new hook

src/components/Messages/index.tsx
```diff
@@ -29,67 +69,63 @@ const Messages: FC = () => {
...
+  if (!isMessagesEnabled) {
     return <Custom404 />;
   }
 
+  if (profilesError) {
+    return <Custom500 />;
+  }
+
+  if (!messagePreviews) {
+    return null;
```
bhavya2611, line 126: We should return something like a loader here, and if absolutely no conversations / messages exist we should have a no message / no conversation UI to be shown

src/components/Messages/index.tsx
```diff
@@ -29,67 +69,63 @@ const Messages: FC = () => {
...
-            setConversations(new Map(conversations));
-          }
+          newConversations.set(convo.peerAddress, convo);
+          return await fetchMostRecentMessage(convo);
         })
-      ).then(() => {
-        setLoading(false);
-      });
-    }
-    const streamConversations = async () => {
```
bhavya2611, line 57: Any reason we removed the streaming of conversations?

We need this when new conversation is started with you, that is already not listed

bhavya2611 (changes requested):

src/components/Messages/Preview.tsx
```diff
@@ -0,0 +1,36 @@
...
+interface Props {
+  profile: Profile;
+  message: Message;
+}
+
+const Preview: FC<Props> = ({ profile, message }) => {
+  const router = useRouter();
+
+  const onConversationSelected = (address: string) => {
+    router.push(address ? `/messages/${address}` : '/messages/');
```
bigint, line 21:
```suggestion
    router.push(address ? `/messages/${address}` : '/messages');
```

src/components/Messages/index.tsx
```diff
@@ -1,26 +1,66 @@
...
-import { useEffect, useState } from 'react';
+import { useEffect } from 'react';
 import { APP_NAME } from 'src/constants';
 import Custom404 from 'src/pages/404';
+import Custom500 from 'src/pages/500';
 import { useAppStore } from 'src/store/app';
 import { useMessageStore } from 'src/store/message';
 import { useSigner } from 'wagmi';
 
+export class MessagePreview {
```
bigint, line 20: **suggestion:** class 😕 we avoid using it 🙈

src/components/Shared/UserProfile.tsx
```diff
@@ -16,18 +16,22 @@ interface Props {
   profile: Profile;
   showBio?: boolean;
   showFollow?: boolean;
+  showHandle?: boolean;
   followStatusLoading?: boolean;
   isFollowing?: boolean;
   isBig?: boolean;
+  messagePreview?: string;
```
bigint, line 23: **nitpick:** looks like a not good idea to add `messagePreview` in this component

Devs won't understand what it does when they are using this component somewhere

bigint (changes requested):

src/components/Messages/index.tsx
elisealix22: Ooooh ok good to know! I thought it might be clean to stitch the Lens & XMTP data together for each item in the list -- but I think I can do it without the new data type 👀

src/components/Messages/index.tsx
elisealix22: Yeah, I should've added this to the PR description, but I removed it for simplicity for now because I wanted to nail down the mapping of the paginated `Profiles` query to the non-paginated convos/messages fetch first. Didn't want another hook potentially updating the state _yet_, but I opened an issue to add this back in a follow-up here & assigned to myself: https://github.com/xmtp-labs/hq/issues/732.

src/components/Messages/index.tsx
elisealix22: 100% agree! I actually had it in this [PR](https://github.com/lensterxyz/lenster/pull/779/commits/52db08a1d885d79e1d7103e614564b4936355f87#diff-5635ed0d7bc771d451cd29937e15f80549c88f309500bde6f80ed7a282bbe5d9L153), but removed it because it was showing when the data was available and I realized it would be easier to add when I pull this out to its own hook.

Tracking here: https://github.com/xmtp-labs/hq/issues/735

src/components/Shared/UserProfile.tsx
elisealix22: Yeah I was going back and forth on creating a new component similar to this (but stripped down) specifically for messages or not. I'll pull it out into its own unless you have other ideas!

src/components/Messages/index.tsx
elisealix22: Agree! Was planning to clean this up in the next PR and keep this one strictly to the data mapping for now because there are still moving pieces in the data flow (paginating profiles and adding streaming back). You cool if we clean up the hook/loading/error state in its own PR? I have issues tracking everything and I agree with all your comments 😆

bhavya2611: Will do

src/components/Shared/UserProfile.tsx
bigint: It's good creating new component for this 😊

elisealix22:
Caught up with @bhavya2611 and we're aligned on the future of the component structure / state handling. I'll be adding a new `ConverstaionList` component to handle most of this state in the next PR.

bhavya2611 (approved):
Looks Good To Me.

src/components/Messages/Message.tsx
```diff
@@ -14,19 +15,17 @@ import { useAppStore } from 'src/store/app';
 import { useMessageStore } from 'src/store/message';
 
 const Message: FC = () => {
-  const { push, query } = useRouter();
-  const address = query.address as string;
+  const { query } = useRouter();
+  const address = (query.address as string).toLowerCase();
```
neekolas, line 19: We should be careful about casing here. The SDK is case-sensitive when looking for contacts and will show someone as not on the network if the casing is wrong (although we have an open issue to always convert to checksum case).

src/graphql/queries/Profiles.graphql
```diff
@@ -0,0 +1,7 @@
+query Profiles($request: ProfileQueryRequest!) {
+  profiles(request: $request) {
+    items {
+      ...ProfileFields
```
neekolas, line 4: Are we sure we need all these fields? Would be nice to pare down to just what is necessary for the preview.

neekolas (approved):

src/components/Messages/Message.tsx
elisealix22: Oooooh great callout! I want to guarantee a match between xmtp addresses and `profile.ownedBy` here since addresses are the only thing stitching the data together. Filed a follow-up to ensure/test case insensitivity: https://github.com/xmtp-labs/hq/issues/736.

src/graphql/queries/Profiles.graphql
elisealix22: Yeah, I think `bio` is the only thing we aren't using currently. We run the risk of the fragment bloating in the future, but gain confidence we'll have enough data to populate any components that depend on Profile. I thought it was an ok tradeoff since it's slim, but will defer to @bigint on his preference 👀 .

```graphql
fragment ProfileFields on Profile {
  id
  name
  handle
  bio
  ownedBy
  isDefault
  attributes {
    key
    value
  }
  picture {
    ... on MediaSet {
      original {
        url
      }
    }
    ... on NftImage {
      uri
    }
  }
  followModule {
    __typename
  }
}
```

src/graphql/fragments/ProfileFields.graphql
```diff
@@ -4,6 +4,7 @@ fragment ProfileFields on Profile {
   handle
   bio
   ownedBy
+  isDefault
```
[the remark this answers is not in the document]
bigint: Also i can't see isDefault is consumed anywhere 🤔

cc @elisealix22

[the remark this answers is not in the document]
elisealix22: >Better we can move isDefault to Profiles.graphql query something like

Love it! Great suggestion. `isDefault` is [used in the PR built on top of this](https://github.com/lensterxyz/lenster/pull/793/files#diff-8c01de88adadf082ed2e906fdcd8d1691afb41cd89156ac62ec1eab281364a0aR50) so I'll remove it from this PR and add it back in #793.

elisealix22:
Alright, should be all set now! Thank you everyone for the reviews!

bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #793, opened 2022-10-11
feat: introduce message preview list component

elisealix22:
## What does this PR do?
This PR builds on top of https://github.com/lensterxyz/lenster/pull/779 and introduces a new `MessagePreviewList` component share logic between the /messages and /message/[addy] pages. It also moves over the state management to a new `useMessagePreviews()` hook and adds a bunch of polish like:
- Paginate through all profiles to ensure we aren't dropping any messages
- Sort the message preview list by most recent message
- Show the most recent message in the message preview text
- Add a loading state in the left panel when message previews are loading (waits for both messages and profiles to complete before rendering)
- Adds back streaming messages

<details>
<summary>Live stream demo 🎥 </summary>

https://user-images.githubusercontent.com/556051/195186758-6768cfd9-5b05-4639-8567-d98f034d38f4.mov
</details>

## Type of change

## How should this be tested?

- [x] Test profile pagination by having more than one conversation started and setting the `limit` to 1 [here locally](https://github.com/lensterxyz/lenster/compare/ea/add-convo-preview...ea/convo-list?expand=1#diff-8c01de88adadf082ed2e906fdcd8d1691afb41cd89156ac62ec1eab281364a0aR30) 
- [x] Test new stream by having Lenster open on messages and sending a new message (see video in description)
- [x] Test loading by ensuring a loading indicator shows up in the left panel on first load


elisealix22:
@bigint Planning to update this base branch to `main` after #779 merges unless you want to just collapse them. Whatever's easiest!

src/components/utils/hooks/useMessagePreviews.tsx
```diff
@@ -0,0 +1,158 @@
...
+      // Paginate through all profiles for the existing conversations.
+      if (pageInfo.next) {
+        await fetchMore({
+          variables: { request: { ...request, cursor: pageInfo?.next } }
+        });
+      }
+    }
+  });
+
+  useEffect(() => {
```
bhavya2611, line 65: I think this should be put into a new hook of its own. The initClient part

src/components/utils/hooks/useMessagePreviews.tsx
```diff
@@ -0,0 +1,158 @@
...
+        limit: 1,
+        direction: SortDirection.SORT_DIRECTION_DESCENDING
+      });
+      if (newMessages.length <= 0) {
+        return { address: peerAddress };
+      }
+      return { address: peerAddress, message: newMessages[0] };
+    };
+
+    const listConversations = async () => {
```
bhavya2611, line 97: Similarly, if we are at a point breaking things down, listingConversations and streamingConversations can also be into two different hooks

bhavya2611 (commented):
Great work, a couple of minor things apart from that all looks good 😄

src/components/utils/hooks/useMessagePreviews.tsx
elisealix22: I like that -- updated!

src/components/utils/hooks/useMessagePreviews.tsx
elisealix22: The streaming/listing here is specific to the message preview list so I'd actually prefer to keep it encapsulated in this hook for now. I don't want it to be misused elsewhere (it only fetches one most recent message) and I enjoy that I can share the `fetchMostRecentMessage` when either the list or stream return to ensure everything is handled the same way.

Feel free to push back if you feel strongly -- otherwise I'd rather leave this as-is for now!

bhavya2611: Ahh makes sense. We can keep it this way

src/lib/cursorBasedPagination.ts
```diff
@@ -31,7 +31,9 @@ export function cursorBasedPagination<T extends CursorBasedPagination>(
           ...pageInfo,
           // reduce total count by excluding dangling items so it won't cause a new page query
           // after item was removed from the cache (for .e.g deleted publication)
-          totalCount: pageInfo?.totalCount - danglingItems?.length
+          totalCount: pageInfo?.totalCount
```
elisealix22, line 34: @bigint after running a `yarn graphql-codegen` totalCount became nullable which was causing a [type check issue](https://github.com/lensterxyz/lenster/actions/runs/3234597297/jobs/5298074546#step:7:8). Want to make sure you see this 👀

bigint: Should be fixed now 😊 can you revert and update the branch 🙇🏼

elisealix22: Appreciate you! Done!

bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #824, opened 2022-10-14
fix: use profileId for message routing

elisealix22:
## What does this PR do?
This PR moves us over to using profileIds to identify conversations. This will be necessary when we move to allowing conversations per-profile instead of per-address.

I also fixed the message body not truncating properly in the preview list. It now shows an ellipsis if the message is long.

## Type of change

## How should this be tested?

- Navigate to each conversation in the message preview list and ensure they load properly.


bigint:
Can we add a loading spinner somewhere till the profile request fulfilled?

But we can make that on another PR 🙇🏼

bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #826, opened 2022-10-14
feat: add enable DMs panel

elisealix22:
## What does this PR do?

This PR adds a call to action to enable DMs. This will help onboard people into the network so they can receive DMs from their friends.

Screen Shot 2022-10-14 at 4 00 31 PM

## Type of change

## How should this be tested?
- Create a new eth/polygon address
- Sign up with Lenster
- Notice the call to action to enable DMs on the homescreen
- Tapping it should take you to messages which will trigger a signature req


bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #823, opened 2022-10-14
chore: Unexpected `await` inside a loop

shanpriyan:
## What does this PR do?

1. Included `Promise.all` 
2. All Promises will run concurrently instead of one after another
3. Fixes - Found `await` inside of loops - [JS-0032](https://deepsource.io/gh/lensterxyz/lenster/issue/JS-0032/occurrences?listindex=0)

Ref: https://www.techiediaries.com/promise-all-map-async-await-example/

## Checklist

- I haven't checked if my changes generate no new warnings
- I haven't added tests that prove my fix is effective or that my feature works
- I haven't checked if new and existing unit tests pass locally with my changes
- I haven't tested the changes in production

PR #827, opened 2022-10-15
Revert "chore: Unexpected `await` inside a loop"

bigint:
Reverts lensterxyz/lenster#823

cc @shanpriyan


shanpriyan:
Is it breaking the way that it should behave?

shanpriyan:
oh, its updated here #828 🚀

PR #828, opened 2022-10-15
fix: upload multiple files at ones

bigint:
@shanpriyan fixed here by converting filelist to js array


shanpriyan:
> @shanpriyan fixed here by converting filelist to js array

oh, cool

Issue #829, opened 2022-10-15
Feedback and findings on DMs


bigint:
- [x] #830
- [x] #831
- [x] #832
- [x] #833
- [x] #834
- [x] #835
- [x] #836
- [ ] #837
- [x] #858
- [x] #857
- [x] #860
- [x] #868
- [x] #870
- [x] #869
- [x] #885
- [x] #892
- [x] #891
- [ ] #909
- [ ] #910
- [x] #911
- [x] #912
- [x] #943
- [ ] #946
- [x] #947
- [x] #948
- [ ] #949
- [x] #971
- [x] #972
- [x] #973
- [x] #989
- [x] #1007
- [x] #1010
- [x] #1013
- [x] #1064
- [x] #1065
- [x] #1066
- [x] #1067
- [x] #1068

bigint:
Closing this

PR #859, opened 2022-10-18
feat: new conversation

bhavya2611:
## What does this PR do?

Adds flow for a new conversation

Fixes # (issue)

Resetting message store and XMTP client on Account Change

## Type of change

## How should this be tested?

- [ ] Try to start a new conversation
- [ ] Change Accounts and see if the message state is refreshed

Default State
image

Search State
image

Profile not on XMTP
image

Profile on XMTP 
image


src/components/Messages/PreviewList.tsx
```diff
@@ -30,13 +32,17 @@ const PreviewList: FC = () => {
...
+  };
+
   return (
     <GridItemFour>
       <Card className="h-[86vh]">
         <div className="flex justify-between items-center p-5 border-b">
           <div className="font-bold">Messages</div>
           <button>
-            <PlusCircleIcon className="h-6 w-6" />
+            <PlusCircleIcon onClick={newMessageClick} className="h-6 w-6" />
```
bigint, line 45: **nitpick:** add onClick to button above

src/components/Messages/PreviewList.tsx
```diff
@@ -30,13 +32,17 @@ const PreviewList: FC = () => {
     return (messageA?.sent?.getTime() || 0) >= (messageB?.sent?.getTime() || 0) ? -1 : 1;
   });
 
+  const newMessageClick = () => {
+    router.push('/messages/');
```
bigint, line 36:
```suggestion
    router.push('/messages');
```

src/components/Messages/index.tsx
```diff
@@ -1,29 +1,64 @@
...
+        setMessageProfiles(new Map(messageProfiles));
+        const newConvo = await client?.conversations?.newConversation(profile.ownedBy);
+        conversations.set(profile.ownedBy.toLowerCase(), newConvo);
+        setConversations(new Map(conversations));
+        router.push(`/messages/${profile.id}`);
+        setError('');
+      }
+    } else {
+      setError('Selected Lens Profile is not on XMTP');
+    }
```
bigint, line 46: **suggestion:** we can avoid else by handling not condition first and return with setError

bigint (changes requested):
@bhavya2611 left some comments

bigint:
Better we can use a modal when a new message button is clicked like Twitter does it

Screenshot 2022-10-18 at 8 05 50 PM

If nothing is selected we can show a state like this

Screenshot 2022-10-18 at 8 06 30 PM

data/feature-flags.ts
```diff
@@ -12,7 +12,8 @@ export const featureFlags = [
       '0x5cce', // saulmc.lens
       '0x014309', // iambhavya.lens
       '0x8539', // petermdenton.lens
-      '0x013f9d' // shanemac.lens
+      '0x013f9d', // shanemac.lens
+      '0x4cc3' // iambhavya2.lens
```
elisealix22, line 16: Is this for testnet? If so, everyone should be enabled by default in testnet. Only mainnet profiles need to be allowlisted here:

https://github.com/lensterxyz/lenster/blob/bf58e32fb178d8e6be3d6035e804fbb77d1c001f/src/lib/isFeatureEnabled.ts#L7

src/components/Shared/Navbar/Search.tsx
```diff
@@ -96,8 +98,14 @@ const Search: FC<Props> = ({ hideDropdown = false }) => {
               <>
                 {profiles.map((profile: Profile) => (
                   <div key={profile?.handle} className="py-2 px-4 hover:bg-gray-100 dark:hover:bg-gray-800">
-                    <Link href={`/u/${profile?.handle}`} onClick={() => setSearchText('')}>
-                      <UserProfile profile={profile} />
+                    <Link
+                      href={isParentMessage ? `` : `/u/${profile?.handle}`}
+                      onClick={() => {
+                        onProfileSelected && onProfileSelected(profile);
```
elisealix22, line 104: Would it be possible to move the previous profile link to an onProfileSelected callback so we don't have competing `href` and `onClick`s and they can both use `onClick()`?

elisealix22 (commented):
Great catch on resetting when the profile changes! Two things on my end:
1. I think we can likely clean up the `Link` vs `onClick` to use a single `onProfileSelected()` callback that defaults to opening a profile if one isn't set.
2. I too was able to duplicate a convo so we'll prob want to protect from that.

After seeing @bigint 's suggestion and chatting with @darickdang, I'm cool with using a modal for searching for new profiles if you are @bhavya2611 ! I think we'd just want to always open the convo after one is selected and eventually (follow-up PR likely) show a warning in that view if they aren't yet on the network:

Screen Shot 2022-10-18 at 4 43 43 PM

src/components/Shared/UserProfile.tsx
```diff
@@ -27,13 +28,14 @@ const UserProfile: FC<Props> = ({
...
-  isBig = false
+  isBig = false,
+  linkToProfile = false
 }) => {
   const [following, setFollowing] = useState(isFollowing);
 
   return (
     <div className="flex justify-between items-center">
-      <Link href={`/u/${profile?.handle}`}>
+      <Link href={linkToProfile ? `/u/${profile?.handle}` : ''}>
```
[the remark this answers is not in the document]
bhavya2611: I think the cursor should still be pointer, as it is clickable, just it will perform different routing and not the one to profile.

bhavya2611:
@bigint @elisealix22 with the time crunch we have do you think UX changes can be taken as enhancement and we for now keep the current flow and keep going? 

The current flow was made as per the designs we had, not that we should not change, but it might be nice to first get all stuff done have a version out and then come back to this. 

Feel free to let me know if you guys feel otherwise :D

bhavya2611:
Also FYI - Apart from the UX change, everything else on this PR is resolved and the PR is good to merge

bigint:
Cool sure thing @bhavya2611 I added to the findings list here https://github.com/lensterxyz/lenster/issues/829

we can check this back later 🙇🏼

PR #866, opened 2022-10-19
Conversation filtering

neekolas:
## What does this PR do?

- Upgrades to beta version of `xmtp-js` SDK
- Uses the new `conversationId` field to filter conversations, which allows for multiple conversations between address pairs
- Updates URL format for messages to now include the full `conversationKey`. This is necessary as there may now be multiple conversations with a given `profileId` (for example, when a profile is transferred to another wallet).
- Gets profiles from Lens API by ID instead of `ownedBy` now that this information is available from the XMTP SDK

NOTE: All existing conversations in the app will disappear with this change. To initiate a new conversation, you can use the message button on a profile to send a 'gm' message which will then make the conversation appear in the list

https://user-images.githubusercontent.com/65710/196741229-a5e3cb9c-85fc-478b-a8a7-b93702fdfd1e.mp4

## Type of change

## How should this be tested?

- [x] Go to a user's profile page (who is also a XMTP user) and click the message button. A message should be sent
- [x] Go to the messages page. All conversations should be visible with previews showing
- [x] Click on a conversation. The most recent 15 messages should appear in the correct order
- [x] Send more than 15 messages in a conversation. Refresh the page. Scroll up. Messages should be added incrementally


src/components/utils/hooks/useMessagePreviews.tsx
```diff
@@ -23,75 +26,79 @@ const useMessagePreviews = () => {
...
   const [messagesLoading, setMessagesLoading] = useState<boolean>();
 
-  const peerAddresses = Array.from(conversations.keys());
-  const request = { ownedBy: peerAddresses, limit: 50 };
-  const {
-    loading: profilesLoading,
-    error: profilesError,
-    fetchMore
-  } = useQuery(ProfilesDocument, {
+  const profileIds = Array.from(conversations.keys())
```
neekolas, line 29: I don't love having to jump through all these hoops to get the `profileIds`. Might replace with a `Set<string>` to store all the `profileIds` discovered

src/components/utils/hooks/useMessagePreviews.tsx
```diff
@@ -23,75 +26,79 @@ const useMessagePreviews = () => {
...
+    .map((conversationKey: string) => {
+      const match = parseConversationKey(conversationKey);
+      return (
+        match && match.members.find((member) => member && currentProfile && member !== currentProfile.id)
+      );
+    })
+    .filter((profileId) => !!profileId);
+
+  const request = { profileIds: profileIds };
```
elisealix22, line 38: I'm not sure I fully understand the motivation to query by profile ids here instead of addresses so I ran thru a profile transferred scenario and here's what I'm thinking:

- We're messaging with a person at ETH address 0x123 who owns profile A & B, and we're talking with A.
- 0x123 transfers profile A to ETH address 0x789.
- We query convos and get our convo history with `0x123-A` back.
- Option 1: We query for profiles by id (A) which now has 0x789 as the owner: `0x789-A` is the profile key.
- Option 2: We query for profiles by address (0x123) which returns profile B: `0x123-B` is the profile key.

Neither option 1 or option 2's new profile key will match a message with the original key `0x123-A` so neither should render as-is right now. Also in either scenario, I think we could look at all profile keys that don't match messages but have either the same profile id (option 1) or eth address (option 2) in the key, and render those convos as locked.

Couldn't querying profiles by addresses still work? Or am I missing something behind the motivation to use profile ids? I think I'm partial to using addresses to stitch xmtp <> lens data together since it's the only constant, but would love to hear your thoughts!

neekolas: One advantage to querying by profile IDs is that we only get back the profiles we actually care about. If someone happens to own 20, it won't bloat the list of returned results when we only care about 1. 

But yes, either way is possible. We could totally query by address and then zip back together the correct profile:convo pairs from the results and throw out the mismatches.

neekolas:
@bhavya2611 @elisealix22 @bigint I'd like to get this merged tomorrow before we run into more merge conflicts. Any objections?

src/components/utils/hooks/useMessagePreviews.tsx
elisealix22: Alright cool, makes sense! I'm good with profileIds then. Thanks for helping walk me thru it. Feel like I understand the flow now.

src/components/utils/hooks/useMessagePreviews.tsx
```diff
@@ -67,43 +69,60 @@ const useMessagePreviews = () => {
...
     };
 
     const listConversations = async () => {
       setMessagesLoading(true);
       const newPreviewMessages = new Map(previewMessages);
       const newConversations = new Map(conversations);
-      const convos = (await client?.conversations?.list()) || [];
+      const newProfileIds = new Set(profileIds);
+      const convos = await client.conversations.list();
+      const matcherRegex = conversationMatchesProfile(currentProfile.id);
```
elisealix22, line 97: Feel free to ignore, but looks we can share this between list & stream methods if we want.

Also, love this way of verifying the convo is for the correct profile while building the convo list! Super neat to see filtering working in action.

src/store/message.ts
```diff
@@ -22,6 +24,22 @@ export const useMessageStore = create<MessageState>((set) => ({
...
   setMessages: (messages) => set(() => ({ messages })),
+  addMessages: (key: string, newMessages: Message[]) => {
+    let numAdded = 0;
+    set((state) => {
+      const messages = new Map(state.messages);
+      const existing = state.messages.get(key) || [];
+      const updated = getUniqueMessages([...existing, ...newMessages]);
+      numAdded = updated.length - existing.length;
+      // If nothing has been added, return the old item to avoid unnecessary refresh
+      if (!numAdded) {
```
elisealix22, line 35: Wait, does zero equate to false here? TIL! Fantastic tweak though 👏 .

elisealix22 (commented):
The code looks great to me from an implementation POV, but I'm having trouble testing. If I go to a profile of someone I've been previously messaging with like `/u/7950test.test` and tap the message button, I can see the message send successfully and render on XMTP.

However, when I go to my inbox, it's empty and I see the following `TypeError` in the console:
Screen Shot 2022-10-19 at 7 15 08 PM

Could it be some collision of a V1 & V2 conversation?

neekolas:
> However, when I go to my inbox, it's empty and I see the following `TypeError` in the console:

@elisealix22 that should be fixed with the upgrade to 7.0.0-beta.2. That was how I spotted the issue with the sha256 function. Maybe you need to do a fresh `yarn` install.

src/components/utils/hooks/useMessagePreviews.tsx
neekolas: Ya, I could make it into a hook or something. Maybe I'll do that as a followup.

bigint: Followup would be better for this 🙏🏼

PR #898, opened 2022-10-21
fix: UI issues

### LINKED ISSUE #891: Send button and the message tile are not aligned in same line
image

### LINKED ISSUE #892: Profile list orders are not getting updated on new messages
https://user-images.githubusercontent.com/69431456/197115770-d68c1025-9ab6-4e68-a493-29828263f1ab.mov

bhavya2611:
## What does this PR do?

Fixes UI Issues

Fixes #892
Fixes #891

## Type of change

## How should this be tested?

- [ ] Test if the alignment of Send button looks good
- [ ] Send a new message and see if the profile list updates
- [ ] Leafwatch events added

Screenshots -

Alignment when the scroll is visible

image

Alignment when the scroll is not visible 

image

Before sending a new message

image

Profile list updates as soon as a new message is sent 

image


bhavya2611:
Note - Let me know if more events need to be tracked I can add @bigint

src/components/Messages/Message.tsx
```diff
@@ -52,12 +55,19 @@ const Message: FC<MessageProps> = ({ conversationKey }) => {
     }
   }, [conversationKey, hasMore, messages, endTime]);
 
+  const showLoading = !missingXmtpAuth && (!profile || !currentProfile || !selectedConversation);
+
+  useEffect(() => {
+    if (profile) {
+      Leafwatch.track('Pageview', { path: PAGEVIEW.MESSAGE });
+      Leafwatch.track(`Conversation with ${profile.ownedBy}`);
```
elisealix22, line 63: I think we should avoid uploading any user info here (even `profile.ownedBy`) if we can avoid it. I'm not sure how useful this info would be and it can lead to privacy or PMI leaks.

bigint:
@bhavya2611 you can revert leafwatch wala part, I'll take care of those 😊

bigint:
@bhavya2611 can you also attach screenshots of changes with before and after 🙏🏼

src/components/Messages/PreviewList.tsx
```diff
@@ -27,8 +27,8 @@ const PreviewList: FC = () => {
   const showLoading = loading && (messages.size === 0 || profiles.size === 0);
 
   const sortedProfiles = Array.from(profiles).sort(([, a], [, b]) => {
-    const messageA = messages.get(a.ownedBy.toLowerCase());
-    const messageB = messages.get(b.ownedBy.toLowerCase());
+    const messageA = messages.get(profileToKey.get(a.id) ?? '');
```
elisealix22, line 30: I don't think we need this new `profileToKey` map in the store. You should be able to use `buildConversationKey` to get the key with the given data and then we don't have to worry about keeping another map in sync.

```js
    const conversationId = buildConversationId(currentProfile.id, profile.id);
    const conversationKey = buildConversationKey(profile.ownedBy, conversationId);
```

elisealix22 (changes requested):
Nice catch on the sorted profiles! Just a couple of comments

elisealix22 (approved):

src/components/Messages/PreviewList.tsx
```diff
@@ -27,8 +29,12 @@ const PreviewList: FC = () => {
   const showLoading = loading && (messages.size === 0 || profiles.size === 0);
 
   const sortedProfiles = Array.from(profiles).sort(([, a], [, b]) => {
-    const messageA = messages.get(a.ownedBy.toLowerCase());
-    const messageB = messages.get(b.ownedBy.toLowerCase());
+    const conversationIdA = buildConversationId(currentProfile?.id, a.id);
```
neekolas, line 32: You don't actually need to rebuild the `conversationKey` here. The key in the `profiles` map is the conversationKey

elisealix22: LOL great catch 🧠 . So you should be able to do

```js
const sortedProfiles = Array.from(profiles).sort(([keyA, a], [keyB, b]) => {
```

bhavya2611: Thanks

bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #900, opened 2022-10-21
chore: remove == null check

### LINKED ISSUE #886: (JS-0059) Null comparisons without type-checking operators may not work as intended
## Description
Comparing to `null` without a type-checking operator (`===` or `!==`), can have unintended results as the comparison will evaluate to `true` when comparing to not just a `null`, but also an `undefined` value.

## Occurrences
There is 1 occurrence of this issue in the repository.

See all occurrences on DeepSource &rarr; [deepsource.io/gh/lensterxyz/lenster/issue/JS-0059/occurrences/](https://deepsource.io/gh/lensterxyz/lenster/issue/JS-0059/occurrences/)

elisealix22:
## What does this PR do?

Use a try/catch to determine whether we want to toast a message failed to send error instead of a null check.

Fixes #886

### DIFF
--- a/src/components/utils/hooks/useSendMessage.tsx
+++ b/src/components/utils/hooks/useSendMessage.tsx
@@ -7,8 +7,12 @@ const useSendMessage = (conversation?: Conversation) => {
       if (!conversation) {
         return false;
       }
-      const sent = await conversation.send(message);
-      return sent != null;
+      try {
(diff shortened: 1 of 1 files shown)


bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #901, opened 2022-10-21
chore: use production messaging env on mainnet

elisealix22:
## What does this PR do?
This PR uses XMTP's `production` environment on mainnet and `dev` on testnet. Note: you'll have to re-sign after this since we're now storing creds per env. Also, mainnet messages will be reset.

## Type of change

## How should this be tested?

- Ensure messages are still visible on testnet after another sig.


bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #902, opened 2022-10-21
fix: insert date divider before date's messages

elisealix22:
## What does this PR do?

This PR updates our UI to insert the date divider when the date changes so that it now shows the messages for the given date below the divider. This matches how other messaging platforms render their dividers like Signal and will likely feel more intuitive.

| Before | After |
|--|--|
| before| after|


bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #923, opened 2022-10-24
feat: add loading and empty state for messaging

elisealix22:
## What does this PR do?
This PR adds a new "Awaiting signature" loading state for the first time the user has to sign to get their XMTP identity key. It also adds an empty state for when they are fully auth'd but haven't started any conversations yet.

| Awaiting signature | Empty | Signed out |
|--|--|--|
| awaiting_centered |  empty_centered | Screen Shot 2022-10-25 at 11 11 56 AM |

## Type of change

## How should this be tested?

- [x] Sign up with a new account that has not enabled DMs yet
- [x] Go to lenster.xyz/messages and notice a new awaiting signature loading state
- [x] Sign the signature request and notice the empty state renders


src/components/utils/hooks/useXmtpClient.tsx
```diff
@@ -29,15 +29,17 @@ const wipeKeys = (walletAddress: string) => {
...
   const currentProfile = useAppStore((state) => state.currentProfile);
   const client = useMessageStore((state) => state.client);
   const setClient = useMessageStore((state) => state.setClient);
+  const [awaitingXmtpAuth, setAwaitingXmtpAuth] = useState<boolean>();
   const isMessagesEnabled = isFeatureEnabled('messages', currentProfile?.id);
 
   useEffect(() => {
     const initXmtpClient = async () => {
       if (signer && !client && currentProfile) {
+        setAwaitingXmtpAuth(true);
```
neekolas, line 42: Think we want this in the `if (!keys) {` block. Otherwise it is going to set the `awaitingXmtpAuth` to true even when the keys are cached in LocalStorage

src/components/Messages/PreviewList.tsx
```diff
@@ -62,17 +65,41 @@ const PreviewList: FC = () => {
...
+              <PageLoading message="Loading conversations" />
+            </div>
+          ) : !currentProfile ? (
+            <div className="flex h-full justify-center items-center">
+              <EmptyState
+                message={<div>Login to start messaging</div>}
+                icon={<MailIcon className="w-8 h-8 text-brand" />}
+                hideCard
+              />
+            </div>
```
bigint, line 94: **suggestion:** this check is not required 🙏🏼 cuz we return 404 in higher level

src/components/utils/hooks/useXmtpClient.tsx
elisealix22: Good catch! Updated 👍

src/components/Messages/PreviewList.tsx
bigint: Hmm but we need to have it at page level because DM UI is protected and should not be visible to public 🙏🏼

elisealix22: Updated the signed out screenshot in the description of the PR to demonstrate. Should be all set!

neekolas (approved):

src/components/Messages/PreviewList.tsx
bigint: Amazing 🚀

PR #930, opened 2022-10-25
Feat/multi prof msg

bhavya2611:
## What does this PR do?

Allow filtering conversations per profile

## Type of change

## How should this be tested?

- [ ] With an address that has multiple lens profiles, try to switch between profiles and see if relevant conversations show up
- [ ] With each profile send and receive messages

## Video -

https://www.loom.com/share/ea31483461fa42f7941e73196d7e4631


src/store/message.ts
```diff
@@ -16,6 +16,9 @@ interface MessageState {
   previewMessages: Map<string, Message>;
   setPreviewMessage: (key: string, message: Message) => void;
   setPreviewMessages: (previewMessages: Map<string, Message>) => void;
+  reset: () => void;
+  selectedProfile: string;
```
elisealix22, line 20: Suggestion: Can we name this `selectedProfileId` so it's not confused as a full profile? Or save the full profile instead of the id -- either one!

src/store/message.ts
```diff
@@ -50,5 +53,17 @@ export const useMessageStore = create<MessageState>((set) => ({
       newPreviewMessages.set(key, message);
       return { previewMessages: newPreviewMessages };
     }),
-  setPreviewMessages: (previewMessages) => set(() => ({ previewMessages }))
+  setPreviewMessages: (previewMessages) => set(() => ({ previewMessages })),
+  selectedProfile: '',
+  setSelectedProfile: (selectedProfile) => set(() => ({ selectedProfile })),
+  reset: () =>
+    set((state) => {
```
elisealix22, line 60: Very neat that we don't have to reset the `client` here since the address doesn't change, however -- do we still reset the client properly when logging out? I think we have a `reset()` inside `useXmtpClient` but want to make sure we don't regress there.

src/components/utils/hooks/useMessagePreviews.tsx
```diff
@@ -65,13 +69,6 @@ const useMessagePreviews = () => {
     }
   });
 
-  const reset = () => {
```
elisealix22, line 68: What's the reason to move this out to the store? For example, I think we no longer reset the loading state now 👀

bhavya2611: I saw this was not needed and was not working correctly, so I moved the reset as a store action. I think we don't really need to reset the loading state, as it gets set as it is on page navigation

src/store/message.ts
bhavya2611: I think we have removed reset from there too. I guess we wipe keys on disconnect and set the client to undefined there

elisealix22 (approved):
Looks good to me!

PR #953, opened 2022-10-26
refactor: dm components

bigint:


src/constants.ts
```diff
@@ -18,8 +18,9 @@ export const DEFAULT_COLLECT_TOKEN = getEnvConfig().defaultCollectToken;
 
 export const IS_MAINNET = API_URL === MAINNET_API_URL;
 
-// XMTP_NETWORK
+// XMTP
 export const XMTP_ENV = IS_MAINNET ? 'production' : 'dev';
+export const XMTP_PREFIX = 'lens.dev/dm';
```
bigint, line 23: @neekolas I think we need to change this 🤔

neekolas: Why?

src/lib/conversationKey.ts
```diff
@@ -13,5 +15,9 @@ export const parseConversationKey = (
 
   const [, peerAddress, memberA, memberB] = Array.from(matches);
 
-  return { peerAddress, members: [memberA, memberB], conversationId: `lens.dev/dm/${memberA}-${memberB}` };
+  return {
+    peerAddress,
+    members: [memberA, memberB],
+    conversationId: `${XMTP_PREFIX}/${memberA}-${memberB}`
```
neekolas, line 21: 👍🏼

PR #978, opened 2022-10-27
chore: allowlist more xmtp testers

elisealix22:
## What does this PR do?
This PR adds a few new XMTP testers as well as a few duplicates for existing testers so we can test switching between multiple profiles.


bhavya2611 (approved):
LG

PR #982, opened 2022-10-28
fix: logout routing issue from other pages

bhavya2611:
## What does this PR do?

Fixes Routing issue after logout from the messages page

## Type of change

## How should this be tested?

- [ ] Logout from the home page and see if it redirects to the home page


src/components/Messages/Message.tsx
```diff
@@ -51,11 +51,11 @@ const Message: FC<MessageProps> = ({ conversationKey }) => {
     }
   }, [conversationKey, hasMore, messages, endTime]);
 
-  if (!currentProfile || !isFeatureEnabled('messages', currentProfile.id)) {
+  if (!isFeatureEnabled('messages', currentProfile?.id)) {
     return <Custom404 />;
   }
 
-  const showLoading = !missingXmtpAuth && (!profile || !currentProfile || !selectedConversation);
+  const showLoading = !missingXmtpAuth && (!profile || !selectedConversation);
```
bigint, line 58: **question:** we need user logged-in check right?

bhavya2611: I think it is a redundant check as the profile will not be defined unless current profile exists, because the profile depends on xmtp client connection which required current profile

bhavya2611: Can add it back for safety anyway

src/components/Messages/Message.tsx
```diff
@@ -51,7 +51,7 @@ const Message: FC<MessageProps> = ({ conversationKey }) => {
     }
   }, [conversationKey, hasMore, messages, endTime]);
 
-  if (!currentProfile || !isFeatureEnabled('messages', currentProfile.id)) {
+  if (!isFeatureEnabled('messages', currentProfile?.id)) {
```
elisealix22, line 54: I personally preferred the previous check because I think we want to 404 there even after the feature flag is removed if current profile doesn't exist. So I think the first version just makes it easier to strip out the feature flag code soon

src/components/Messages/PreviewList.tsx
```diff
@@ -32,7 +32,7 @@ const PreviewList: FC<Props> = ({ className }) => {
   const [showSearchModal, setShowSearchModal] = useState(false);
   const { authenticating, loading, messages, profiles, profilesError } = useMessagePreviews();
 
-  if (!currentProfile || !isFeatureEnabled('messages', currentProfile.id)) {
+  if (!isFeatureEnabled('messages', currentProfile?.id)) {
```
elisealix22, line 35: Same here -- I think we still want to 404 here if current profile is null even after the feature flag is stripped so I think the first check might be safer.

src/components/Messages/index.tsx
```diff
@@ -12,7 +12,7 @@ import PreviewList from './PreviewList';
 const Messages: NextPage = () => {
   const currentProfile = useAppStore((state) => state.currentProfile);
 
-  if (!currentProfile || !isFeatureEnabled('messages', currentProfile.id)) {
+  if (!isFeatureEnabled('messages', currentProfile?.id)) {
```
elisealix22, line 15: Last one, but same as above. Please lmk if you disagree or if I'm missing something, but I think it's just safer to explicitly check.

bigint, line 15: we need to have currentProfile check, else it will not render 404 if profile is not there 🙈

elisealix22 (approved):
I'd prefer to explicitly check if currentProfile is around to 404 as noted lol, but this works!

src/components/Messages/index.tsx
bigint: +1 we need this @bhavya2611 🙏🏼

src/components/Messages/Message.tsx
bhavya2611: This one after disconnect will first show 404 and then route to home, if that is okay, I can add it back

bigint: Yup its expected one

PR #1009, opened 2022-10-28
fix: message list shouldn't get stuck loading

elisealix22:
## What does this PR do?
This PR makes it so our `useGetConversation` hook listens for updates to `conversations` since it sets the selectedConversation based on the data in that store. This should fix sometimes there being a null selectedConversation which appeared like an endless loading state.

It also polishes the missing xmtp auth UI for people who haven't enabled DMs yet.

| Before | After |
|--|--|
| CleanShot 2022-10-28 at 14 05 25@2x |  Screen Shot 2022-10-28 at 4 03 33 PM |

## Type of change

## How should this be tested?

- [x] Open a profile of someone who has messages enabled but you haven't DM'd with yet from their profile. Ensure the profile loads.
- [x] Navigate to a profile who hasn't authenticated w XMTP and ensure you can't type a message yet


data/feature-flags.ts
```diff
@@ -27,7 +27,8 @@ export const featureFlags = [
       '0x016b25', // adamalix.lens
       '0x016b24', // bhavyam.lens
       '0x016b0b', // xmtplabs.lens
-      '0xf5ff' // anoopr.lens
+      '0xf5ff', // anoopr.lens
+      '0x016b2e' // kenziestokes.lens
```
elisealix22, line 31: This is another XMTP employee btw!

src/components/Messages/Message.tsx
```diff
@@ -77,7 +77,7 @@ const Message: FC<MessageProps> = ({ conversationKey }) => {
                 hasMore={hasMore}
                 missingXmtpAuth={missingXmtpAuth ?? false}
               />
-              <Composer sendMessage={sendMessage} />
+              <Composer sendMessage={sendMessage} missingXmtpAuth={missingXmtpAuth ?? false} />
```
neekolas, line 80: I would name this something more generic like "disabled", so that the component is more reusable

neekolas (approved):

src/components/Messages/Message.tsx
elisealix22: Dig it! Updated.

src/components/utils/hooks/useGetConversation.tsx
```diff
@@ -44,7 +44,7 @@ const useGetConversation = (conversationKey: string, profile?: Profile) => {
     };
     createNewConversation();
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [profile]);
+  }, [profile, conversations]);
```
neekolas, line 47: This is going to re-run any time any conversation changes. Should probably be scoped to the single conversation we care about.

bigint (approved):

bigint:
@elisealix22 can you please resolve the conflicts 🙏🏼

src/components/utils/hooks/useGetConversation.tsx
elisealix22: Glad you called this out! Updated to only listen for the `selectedConversation`!

elisealix22:
Resolved @bigint ! Sorry for the delay

neekolas (approved):

PR #1036, opened 2022-10-31
feat: redirect to first conversation in list

neekolas:
## What does this PR do?

- Attempts to redirect to the first conversation in the list
- Conditions to allow a redirect are
1. The user's viewport is 1024px or greater (we don't want this behaviour on mobile, since we use the `/messages` as a distinct route to show the conversation list)
2. The user is on the `/messages` route
3. There is at least one conversation in the list loaded.

I think this feature is going to get picked up in a planned refactor to merge the `Message.tsx` and `index.tsx` pages into a single route. Right now, the necessary information to execute this redirect is lower in the React tree than is ideal. 

I also had to add a new React hook to track the window width to avoid this behaviour on mobile, which is unfortunate. Open to other solutions that don't rely on `window.innerWidth`.

The PR also changes the send button to only icon in the mobile view and makes the mobile view flow start from 1024px

Fixes # (issue)

https://user-images.githubusercontent.com/65710/199074880-b8df72ff-104b-45ec-9e87-af2e64cf9976.mp4

image

## Type of change

## How should this be tested?

- [ ] Go to the `/messages` page on the desktop. The first conversation should be automatically selected once loading is complete
- [ ] Go to the `/messages` page on mobile. The conversation list should appear, and it should be possible to navigate to the individual conversation and back.
- [ ] In mobile view the send button should only have an icon

### DIFF
--- a/src/components/Messages/PreviewList.tsx
+++ b/src/components/Messages/PreviewList.tsx
@@ -43,7 +43,6 @@ const PreviewList: FC<Props> = ({ className }) => {
   });
 
   useEffect(() => {
-    console.log(width);
     // Ignore this hook on mobile, since we use the /messages route to show the conversation list
     if (!width || width < MIN_WIDTH_FOR_TWO_PANELS) {
       return;
--- a/src/components/Messages/PreviewList.tsx
+++ b/src/components/Messages/PreviewList.tsx
@@ -16,12 +16,11 @@ import isFeatureEnabled from '@lib/isFeatureEnabled';
 import { useRouter } from 'next/router';
 import type { FC } from 'react';
 import { useEffect, useState } from 'react';
+import { MIN_WIDTH_DESKTOP } from 'src/constants';
 import Custom404 from 'src/pages/404';
 import Custom500 from 'src/pages/500';
 import { useAppStore } from 'src/store/app';
 import { useMessageStore } from 'src/store/message';
-
-const MIN_WIDTH_FOR_TWO_PANELS = 1024;
 
 interface Props {
   className?: string;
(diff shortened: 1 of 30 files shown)


src/components/Messages/PreviewList.tsx
```diff
@@ -7,19 +7,22 @@ import { GridItemFour } from '@components/UI/GridLayout';
...
 import { useRouter } from 'next/router';
 import type { FC } from 'react';
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import Custom404 from 'src/pages/404';
 import Custom500 from 'src/pages/500';
 import { useAppStore } from 'src/store/app';
 import { useMessageStore } from 'src/store/message';
 
+const MIN_WIDTH_FOR_TWO_PANELS = 1024;
```
bhavya2611, line 24: I think the mobile flow starts 768px, we should have that here

bhavya2611 (approved):
Small comment

src/components/Messages/PreviewList.tsx
neekolas: So, maybe it should start at 768. RIght now, between 768 and 1024 it doesn't hide the conversation list when you have a conversation loaded. I did make a separate ticket for that. https://github.com/xmtp-labs/hq/issues/820

bhavya2611: Also move the constant to the constants file

neekolas: Ya, that's a good call

bhavya2611: Okay so are you suggesting we change to a mobile view below 1024px?

I think that will make sense.

bigint:
Can you please resolve the conflicts 🙏🏼

bhavya2611:
@bigint resolved conflicts

src/components/utils/hooks/useWindowSize.tsx
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useState } from 'react';
+
+interface Size {
+  width: number | undefined;
+  height: number | undefined;
+}
+
+// Borrowed from https://usehooks.com/useWindowSize/
+const useWindowSize = (): Size => {
```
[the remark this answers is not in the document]
neekolas: For styling, absolutely. But this is used to change behavior

PR #1040, opened 2022-10-31
fix: search profiles modal width

bhavya2611:
## What does this PR do?

Fixes the width of user profile lists in the search modal

## Type of change

## How should this be tested?

- [ ] See if the width of the list looks good and matches the search text input above

## Screenshot

image


bigint:
This looks very huge in the navbar search 😕 

Lotta whitespace

Screenshot 2022-11-01 at 11 05 04 AM

bhavya2611:
@bigint I added a prop to pass className so we can configure it as required in different places. I hope this works

bigint (changes requested):

bigint (approved):

PR #1047, opened 2022-10-31
feat: add a very basic messaging badging

elisealix22:
## What does this PR do?

Adds a very basic badging system to messages to indicate there is a new message when you are not viewing the tab. It handles switching between profiles and will persist the badge state for each signed in profile until localstorage is cleared or the user signs out.

https://user-images.githubusercontent.com/556051/199167506-310330fc-f44a-4e27-b352-fd9078439490.mov

## Type of change

## How should this be tested?

- [x] Send messages to a profile while not on the messages tab and ensure you get a badge
- [x] Send messages to a profile while viewing the messages tab and ensure you _do not_ get a badge
- [x] Sign out / sign back in and ensure you need a new sig
- [x] Create a new account and ensure it does not prompt for a signature or badge


src/components/Messages/MessageIcon.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const fetchShowBadge = async () => {
+      const convos = await cachedClient.conversations.list();
+      const matchingConvos = convos.filter(
+        (convo) => convo.context?.conversationId && matcherRegex.test(convo.context.conversationId)
+      );
+      if (matchingConvos.length <= 0) {
+        return;
+      }
+
+      const previews = await Promise.all(matchingConvos.map(fetchMostRecentMessage));
```
neekolas, line 44: The one thing that's a little scary here is the number of API requests involved on every pageview. If you have 100 people who have ever messaged you, this will lead to 101 API requests just to show the badge. 

To just get the timestamp of the most recent message, the more performant option would be to do something like.

```ts
const topics = matchingConvos.map((convo) => convo.topic)
const recentMessage = client.listEnvelopes([], (env) => env.timestampNs, { limit: 1, direction: SortDirection.SORT_DIRECTION_DESCENDING})
const lastMessageTimestamp = recentMessages.length ? somehowConvertNanosecondTimestampToDate(recentMessages[0]) : null
```

elisealix22: Nick!!! This is brilliant. Yeah, I wrote down today that we'll likely soon want to be able to paginate & sort by updatedAt for convos 😅 (no idea at how feasible that would be today), but this is so much better for this use case. Updating!

bigint:
CI + Deployment is failing 😕

elisealix22:
@bigint all good! it's failing right now because I'm pointed to a local `@xmtp/xmtp-js` instance until I can deploy a new endpoint: https://github.com/xmtp/xmtp-js/pull/205. Other than that, I'm feeling really quite excited about this one! It's persisting per-profile now too 💾

src/components/Messages/MessageIcon.tsx
```diff
@@ -0,0 +1,120 @@
+import useXmtpClient from '@components/utils/hooks/useXmtpClient';
```
bhavya2611, line 1: We might want to name this component better. MessageIcon component having the code for streaming messages is a bit out of context

src/components/Notification/Icon.tsx
```diff
@@ -34,7 +34,7 @@ const NotificationIcon: FC = () => {
       }}
     >
       <LightningBoltIcon className="w-5 h-5 sm:w-6 sm:h-6" />
-      {showBadge && <span className="w-2 h-2 bg-red-500 rounded-full" />}
+      <span className={`w-2 h-2 bg-red-500 rounded-full ${showBadge ? 'visible' : 'invisible'}`} />
```
bhavya2611, line 37: I think we should use block and hidden here, as visible and invisible will always render the element and just not allocate the space vs it is better to not render the element at all if not required

bhavya2611 (commented):
Small comments

src/components/Notification/Icon.tsx
elisealix22: Amazing suggestion! Will update!

elisealix22: I reverted for now because I couldn't get the desired effect of the icon not moving if the badge comes in. We will have to address that in a separate PR.

src/components/Messages/PreviewList.tsx
```diff
@@ -44,12 +45,18 @@ const PreviewList: FC<Props> = ({ className, selectedConversationKey }) => {
...
   const showLoading = loading && (messages.size === 0 || profiles.size === 0);
 
+  let hasMessages = false;
   const sortedProfiles = Array.from(profiles).sort(([keyA], [keyB]) => {
-    const messageA = messages.get(keyA);
-    const messageB = messages.get(keyB);
-    return (messageA?.sent?.getTime() || 0) >= (messageB?.sent?.getTime() || 0) ? -1 : 1;
+    const aTime = messages.get(keyA)?.sent.getTime() || 0;
+    const bTime = messages.get(keyB)?.sent.getTime() || 0;
+    hasMessages = hasMessages || aTime > 0 || bTime > 0;
```
neekolas, line 52: It's a little confusing having side-effects inside a sort function. Can we move this check elsewhere?

src/components/Messages/PreviewList.tsx
```diff
@@ -44,12 +45,18 @@ const PreviewList: FC<Props> = ({ className, selectedConversationKey }) => {
...
+    return aTime >= bTime ? -1 : 1;
   });
 
+  if (hasMessages) {
+    clearMessagesBadge(currentProfile.id);
```
neekolas, line 57: We should wrap this in a useEffect. Having things that mutate the state inside the render function can lead to some funky behaviour like infinite render loops.

bigint (approved):

PR #1062, opened 2022-11-02
fix: resolve issue where badge would pop back up on messages

elisealix22:
## What does this PR do?

This PR resolves an issue where we were listening for changes to `pathname` which would attempt to return the stream and start a new one. The issue is that as we bounce around different tabs, the async generator would return but it didn't fully close until it yielded something -- which meant an old stream from the notifications tab could receive one last message thinking it was on the right tab to show the message.

The fix is to validate the profile & pathname are valid at runtime when we receive that last message and ignore the message if we are currently viewing the `/messages` tab and signed in to the correct user.

<details>
<summary>Demo of the issue</summary>

https://user-images.githubusercontent.com/556051/199401126-8d7cf16e-4bbf-442b-8ba6-577a815da5a9.mp4

</details>

## How should this be tested?

- [x] Navigate between home, explore & notifications then view messages and receive a message. Notice the badge no longer appears.
- [x] Toggle between two profiles on Home where one has a message and one doesn't and notice it persists.
- [x] I retested the create new user flow as well and ensured no badges.


src/components/utils/hooks/useMessagePreviews.tsx
```diff
@@ -60,9 +60,6 @@ const useMessagePreviews = () => {
         newMessageProfiles.set(key, profile);
       }
       setMessageProfiles(newMessageProfiles);
-
-      // TODO: Add pagination of results once Lens fixes an issue with their API where pagination and cursor
```
elisealix22, line 64: In chatting with Nick & Josh, this is running as intended so I removed the TODO.

src/store/message.ts
```diff
@@ -95,18 +95,14 @@ export const useMessagePersistStore = create(
...
-            const show = new Map(state.showMessagesBadge);
-            viewedAt.set(profileId, toNanoString(new Date()));
-            show.set(profileId, false);
-            return { viewedMessagesAtNs: viewedAt, showMessagesBadge: show };
-          } else {
-            return {
-              viewedMessagesAtNs: state.viewedMessagesAtNs,
-              showMessagesBadge: state.showMessagesBadge
-            };
+          const viewedAt = new Map(state.viewedMessagesAtNs);
```
elisealix22, line 98: I modified this to always update the last viewed timestamp when we call clear()

bigint (approved):

PR #1087, opened 2022-11-03
feat: add shop button to collect on the secondary marketplace

ahomentc:
Add "Shop Collects" button to Lenster.

## What does this PR do?
Shop button will allow users to find a publication's collects on the secondary marketplace. If a publication is sold out of collects, users might want to buy one from secondary. This makes it easy for them to find those collects they want to buy.

Instead of browsing on LensPort for collects they can also browse on Lenster now.

image

## Type of change


bigint:
Moved to https://github.com/lensterxyz/lenster/pull/1099

PR #1089, opened 2022-11-03
fix: paginate profiles when fetching message previews

elisealix22:
## What does this PR do?

Currently, the [get by ids](https://docs.lens.xyz/docs/get-profiles#get-by-profile-ids) Lens query does not support pagination. This PR updates us to use the [get by ownedBy](https://docs.lens.xyz/docs/get-profiles#get-by-owned-by) and paginate thru all profiles that could match conversations.

Fixes Stani not being able to use the messaging feature fully as he has many conversations 😛 .

## Type of change

## How should this be tested?

- [x] Set the limit of the request to 1 and ensure all your conversations still load

### DIFF
--- a/src/components/utils/hooks/useMessagePreviews.tsx
+++ b/src/components/utils/hooks/useMessagePreviews.tsx
@@ -37,7 +37,7 @@ const useMessagePreviews = () => {
       return null;
     }
 
-    return parsed.members.find((member) => member !== userAddress) ?? null;
+    return parsed.peerAddress;
   };
 
   const request = { ownedBy: Array.from(peerAddresses.values()), limit: 50 };
--- a/src/components/utils/hooks/useMessagePreviews.tsx
+++ b/src/components/utils/hooks/useMessagePreviews.tsx
@@ -40,7 +40,7 @@ const useMessagePreviews = () => {
     return parsed.members.find((member) => member !== userAddress) ?? null;
   };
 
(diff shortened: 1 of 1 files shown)


src/components/utils/hooks/useMessagePreviews.tsx
```diff
@@ -27,29 +27,34 @@ const useMessagePreviews = () => {
   const setPreviewMessage = useMessageStore((state) => state.setPreviewMessage);
   const reset = useMessageStore((state) => state.reset);
   const { client, loading: creatingXmtpClient } = useXmtpClient();
-  const [profileIds, setProfileIds] = useState<Set<string>>(new Set<string>());
```
neekolas, line 30: Aren't we going to need both? Otherwise we are going to be adding profiles that the user isn't actually talking to to the `messageProfiles` map

elisealix22: We will add them, but we shouldn't show them right? Since the key doesn't match if the profile id returned is different?

neekolas: Depends. If we iterate through that list of keys anywhere (like we do on the PreviewList) we are going to run into items that exist in the `messagePreviews` but where there are no messages.

In that case, we ignore them. But we'd have to make sure there aren't more cases to watch out for.

PR #1091, opened 2022-11-03
fix: break up profiles into batches of 50

neekolas:
## What does this PR do?

- We realized the Lens API doesn't allow you to query for more than 50 profiles at a time, but also doesn't offer pagination when querying by `profileIds`. So, we are breaking the request into smaller chunks.
- Bonus is that when a new conversation is added it will only query for the one new profile instead of the whole batch

Fixes # (issue)

https://user-images.githubusercontent.com/65710/199830468-ee5fe5e7-8ad9-4cd4-a9cb-c05c1c269ecb.mp4

## Type of change

## How should this be tested?

- [ ] Have > 50 different conversations open. All should load. (More easily tested by lowering the `MAX_PROFILES_TO_LOAD_CONCURRENTLY` variable)

### DIFF
--- a/src/components/utils/hooks/useMessagePreviews.tsx
+++ b/src/components/utils/hooks/useMessagePreviews.tsx
@@ -76,7 +76,7 @@ const useMessagePreviews = () => {
         );
 
         for (const result of results) {
-          if (!result.data?.profiles.items) {
+          if (!result.data?.profiles.items.length) {
             continue;
           }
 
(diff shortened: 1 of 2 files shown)


src/components/utils/hooks/useMessagePreviews.tsx
```diff
@@ -40,26 +46,59 @@ const useMessagePreviews = () => {
...
-        const key = buildConversationKey(peerAddress, buildConversationId(currentProfile?.id, profile.id));
-        newMessageProfiles.set(key, profile);
+      const chunks = chunkArray(Array.from(toQuery), MAX_PROFILES_PER_REQUEST);
+      try {
+        const results = await Promise.all(
+          chunks.map((profileIdChunk) =>
+            apolloClient.query({
+              query: ProfilesDocument,
+              variables: { request: { profileIds: profileIdChunk } }
+            })
```
bigint, line 73: **question:** I think we can use `useLazyQuery` hook here 🤔, if we use `apolloClient` all the apollo config will not be applied here

bigint:
Merging this now, cuz its breaking prod 🙏🏼

PR #1108, opened 2022-11-06
Login

kashyapjii:
Enable single click login for 24 hours
Everytime i try to login in lensters it ask for sign in


Itsdamntrue:
Lets add some donation badge where users add currency for donate someone

PR #1119, opened 2022-11-07
changes to filter conversations as followed and requested

bhavya2611:
## What does this PR do?

Adds the following and requested tab in the conversations preview list to avoid spam.

## Type of change

- [ ] Enhancement (small non-breaking changes to existing functionality)

## How should this be tested?

- [ ] Check if the conversations from the profiles you follow and you don't are displayed in the correct tabs
- [ ] Try to switch between tabs
- [ ] Select a conversation and message them
- [ ] Search a profile and from the profiles, page click on the message button and see if routing is good

## Screenshots

image

image

### DIFF
--- a/src/components/Messages/PreviewList.tsx
+++ b/src/components/Messages/PreviewList.tsx
@@ -9,7 +9,7 @@ import { Modal } from '@components/UI/Modal';
 import { PageLoading } from '@components/UI/PageLoading';
 import useMessagePreviews from '@components/utils/hooks/useMessagePreviews';
 import type { Profile } from '@generated/types';
-import { MailIcon, PlusCircleIcon } from '@heroicons/react/outline';
+import { MailIcon, PlusCircleIcon, UsersIcon } from '@heroicons/react/outline';
 import buildConversationId from '@lib/buildConversationId';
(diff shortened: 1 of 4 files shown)


bigint:
I'm not seeing any DMs in following tab 🤔 but I have lotta DMs from people I follow 🤔

elisealix22:
@bhavya2611 oooh i found one more case that we prob want to handle:

if i tap the "follow" button on a person in the requested tab, i'd expect that convo to move over to "following". also prob vice-versa where if i "unfollow" they move to the other side. would this be v difficult?

neekolas:
@bigint did the latest round of changes fix your issue? I am seeing profiles correctly partitioned in the preview deployment.

bhavya2611:
@elisealix22 can be done, but the problem is that the follow, unfollow state to reflect takes time, I can on the basis of that it went through can move a profile over, but if the state refreshes due to any reason the conversation will again move to the previous tab, till the profile object finally reflects it.

elisealix22 (approved):
✨ Love it!

neekolas (approved):

PR #1127, opened 2022-11-07
fix: ensure correct requested tab count

elisealix22:
## What does this PR do?
This PR builds on https://github.com/lensterxyz/lenster/pull/1119 with a few of fixes:
  1. We ensure we have a message for each convo before incrementing the "Requested" count. Before if you opened a profile to message, it would start a new convo and increment the count even though nothing was visible.
  2. We only loop through the array of profiles once and partition into followed / requested groups.
  3. It updates the selected tab for a new conversation with a new profile. So if you were to kick off a new convo with someone you follow, it would switch to the following tab. Alternatively, if you want to message someone you don't follow, it will select the Requested tab. This ensures the convo will appear in the preview list when the first message is sent because the correct tab will be selected.


src/components/Messages/PreviewList.tsx
```diff
@@ -30,51 +30,30 @@ interface Props {
...
     if (!currentProfile) {
       return;
     }
-    const profileKeys = Array.from(profiles.keys());
-    const messageKeys = Array.from(messages.keys());
-    const hasPreviews = profileKeys.some((item) => messageKeys.includes(item));
-    if (hasPreviews) {
-      clearMessagesBadge(currentProfile.id);
-    }
+    clearMessagesBadge(currentProfile.id);
```
elisealix22, line 51: It should be ok to clear the badge here aggressively since we clear on click of the message icon as well.

src/components/Messages/PreviewList.tsx
```diff
@@ -125,16 +103,16 @@ const PreviewList: FC<Props> = ({ className, selectedConversationKey }) => {
...
+              <span className="text-sm font-bold ml-2 bg-gray-200 px-3 py-0.5 rounded-2xl">
+                {requestedCount > 99 ? '99+' : requestedCount}
+              </span>
+            )}
           </div>
         </div>
         {selectedTab === 'Requested' ? (
           <div className="p-2 mt-1 text-sm bg-yellow-100 text-yellow-800">
-            These conversations are from Lenster profiles that you don't currently follow.
+            These conversations are from Lens profiles that you don't currently follow.
```
elisealix22, line 115: + @darickdang @bigint I tweaked the copy since profiles aren't a Lenster-specific thing. Lmk if you want me to revert!

src/store/message.ts
```diff
@@ -6,7 +6,7 @@ import { LS_KEYS } from 'src/constants';
 import create from 'zustand';
 import { persist } from 'zustand/middleware';
 
-type tabValues = 'Following' | 'Requested';
+type TabValues = 'Following' | 'Requested';
```
elisealix22, line 9: It looked like most types start with a capital letter in the project so I bumped it!

src/store/message.ts
```diff
@@ -61,6 +62,18 @@ export const useMessageStore = create<MessageState>((set) => ({
   },
   messageProfiles: new Map(),
   setMessageProfiles: (messageProfiles) => set(() => ({ messageProfiles })),
+  addProfileAndSelectTab: (key, profile) =>
+    set((state) => {
+      let profiles: Map<string, Profile>;
+      if (!state.messageProfiles.get(key)) {
```
elisealix22, line 68: I tried to be smart here and only trigger a profile refresh if it's a new profile. This is also where we update the selected tab before opening the message component.

bhavya2611 (approved):
LGTM

bigint (approved):

PR #1128, opened 2022-11-08
chore: add appVersion to xmtp client config

saulmc:
## What does this PR do?

Configures the XMTP client to include the app name and version in XMTP API headers.

## Type of change

## How should this be tested?

- [ ] Confirm that headers include `'X-App-Version': 'Lenster/1.0.3-beta'`

### DIFF
--- a/src/components/utils/hooks/useXmtpClient.tsx
+++ b/src/components/utils/hooks/useXmtpClient.tsx
@@ -1,6 +1,6 @@
 import { Client } from '@xmtp/xmtp-js';
 import { useCallback, useEffect, useState } from 'react';
-import { LS_KEYS, XMTP_ENV } from 'src/constants';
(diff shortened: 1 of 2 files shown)


src/components/utils/hooks/useXmtpClient.tsx
```diff
@@ -47,7 +47,11 @@ const useXmtpClient = (cacheOnly = false) => {
           storeKeys(await signer.getAddress(), keys);
         }
 
-        const xmtp = await Client.create(null, { env: XMTP_ENV, privateKeyOverride: keys });
+        const xmtp = await Client.create(null, {
```
neekolas, line 50: You probably want the appVersion on the getKeys line above as well. That actually does do an API request to download the key bundle, so we should track it along with the rest.

bigint (approved):

PR #1179, opened 2022-11-11
fix: url matching

### LINKED ISSUE #771: Fix external link regex in single publication
https://lenster.xyz/posts/0x24-0xc7

foolo:
## What does this PR do?

Fixes #771

Details:
- Fixed ordering of matchers: URLMatcher now goes first to catch any URLs before the other matchers check for mention-tags, hashtags, etc.
- URLMatcher now marked as "void" (which means an URL cannot contain nested tags, see https://interweave.dev/docs/matchers/#creating-a-matcher )
- URLMatcher: when a blocked top-level-domain is found (i.e. ".lens"), set `response.valid = false;` This is the method in Interweave to filter out such false positives. (Before the fix, **null** was returned, which stopped the matcher from finding more matches, see https://interweave.dev/docs/matchers/#creating-a-matcher ) 

Before:
before1

After:
after1

## Type of change

## How should this be tested?

Open  https://lenster.xyz/posts/0x24-0xc7 and check that URL is rendered correctly (whole URL clickable, and no mention-tag inside URL)

### DIFF
--- a/src/components/Shared/Markup.tsx
+++ b/src/components/Shared/Markup.tsx
@@ -19,6 +19,7 @@ interface Props {
 
 const Markup: FC<Props> = ({ children, className = '', matchOnlyUrl }) => {
   const defaultMatchers = [
+    new UrlMatcher('url'),
     new HashtagMatcher('hashtag'),
     new MentionMatcher('mention'),
     new MDBoldMatcher('mdBold'),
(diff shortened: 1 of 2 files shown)


bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #1181, opened 2022-11-11
fix: load profiles serially

neekolas:
## What does this PR do?

There seems to be an issue loading the profiles in parallel, where it uses the same variables for each request. This works around that issue by loading them one at a time. Confirmed it works by lowering the number of concurrent profiles and testing in dev.

Will work on a subsequent fix that loads everything in parallel.

Fixes # (issue)

## Type of change

## How should this be tested?

- [ ] Change the `MAX_PROFILES_PER_REQUEST` to 2 and refresh with a profile that has > 2 ongoing conversations. It should return all the profiles


elisealix22 (approved):
Thanks Nick 💚

bigint (approved):

Issue #123, opened 2022-07-24
(SH-2034) Unused variable


bigint:
## Description
Variables that are declared but not used for anything should be removed.

## Occurrences
There are 2 occurrences of this issue in the repository.

See all occurrences on DeepSource &rarr; [deepsource.io/gh/lensterxyz/lenster/issue/SH-2034/occurrences/](https://deepsource.io/gh/lensterxyz/lenster/issue/SH-2034/occurrences/)

PR #1184, opened 2022-11-12
fix: hashtag matcher

foolo:
## What does this PR do?

- Detect hashtags with digits, for example `#web3`,
- Detect hashtags with underscore, for example `#foo_bar`
- **Ignore** hashtags with only digits and/or underscore, for example `#123`, `#12_34`, `#_`)

Before fix:
before
After fix:
after

Here is a live example post which contains `#Web3` hashtag: https://lenster.xyz/posts/0x73a4-0x029e

## Type of change

## How should this be tested?

- Create and view a post with the text. `#lens #web3 #foo_bar #123 #_ #12_34 #foo! #bar? #test. dfg`
- Verify that only `#lens`, `#web3`, `#foo_bar`, `#foo` and `#bar` are detected as hashtags.

### DIFF
--- a/src/components/utils/matchers/HashtagMatcher.tsx
+++ b/src/components/utils/matchers/HashtagMatcher.tsx
@@ -46,7 +46,7 @@ export class HashtagMatcher extends Matcher {
   }
 
   match(value: string) {
-    return this.doMatch(value, /\B(#[A-Za-z]+\b)(?!;)/, (matches) => {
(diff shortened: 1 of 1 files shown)


foolo:
@bigint I'm not sure if it's OK to just open a PR like this, without creating an issue first. Is it fine?, or should there always be an issue first?

bigint:
It's perfectly fine 😊

bigint (approved):

foolo:
Great, good to know! :)

PR #1188, opened 2022-11-13
feat: lexical markdown render

bigint:
- [x] Render mention slug
- [x] Open slug and mention in the same tab
- [x] Open the external link in a new tab


src/components/Composer/Editor/index.tsx
```diff
@@ -13,7 +13,7 @@ import { usePublicationStore } from 'src/store/publication';
 
 import MentionsPlugin from '../../Shared/Lexical/Plugins/AtMentionsPlugin';
 
-const TRANSFORMERS = [...TEXT_FORMAT_TRANSFORMERS];
+const TRANSFORMERS = [...TEXT_FORMAT_TRANSFORMERS, LINK];
```
harish-sethuraman, line 16: @bigint going forward we have to decide if we want to convert all links to markdown link node or not `[]()`. If we do convert to link node we won't need custom transformer while preview. If we dont add this we can actually add link regex to preview transformer and convert it to link (Link ive done)

src/components/Shared/Markdown/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  ...TEXT_FORMAT_TRANSFORMERS,
+  {
+    dependencies: [MentionNode],
+    export: () => {
+      return null;
+    },
+    regExp: AtSignMentionsRegex,
+    replace: (parentNode: ElementNode, children: Array<LexicalNode>, match: Array<String>) => {
+      // we have to check if the match is a valid mention and then convert it to mentionNode
+      const node = $createMentionNode(match[0].split('@')[1]);
```
harish-sethuraman, line 28: @bigint we have to check if all mentions are a valid mention and then convert it to mention node. Add proper logic here?

src/components/Shared/Markdown/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  {
+    dependencies: [LinkNode],
+    export: () => {
+      return null;
+    },
+    importRegExp: /\[([^[]+)]\(([^(]+)\)/,
+    regExp: URL_MATCHER,
+    replace: (textNode: ElementNode, match: Array<string>) => {
+      const [, linkText, linkUrl] = match;
+      const linkNode = $createLinkNode(linkUrl, {
```
harish-sethuraman, line 44: This is the place for preview transformer where we test for link regex and then convert it to link node.

src/components/Composer/Editor/index.tsx
bigint: Actually we need both, some apps in lens ecosystem will just post link some post markdown link, so we need to handle all edge cases 🙇🏼

src/components/Shared/Markdown/index.tsx
bigint: Hmm I think we can better catch mentions by regex 🤔 

If we use this it will break in many places, for example TikTok URL will look something like this https://tt.com/@strek/493849385983

harish-sethuraman: No No test the regex with the link. It wont detect this as mention. 

Screenshot 2022-11-14 at 11 54 22 AM

bigint: cool then 🚀

src/components/Shared/Markdown/PreviewMentionNode.tsx
```diff
@@ -0,0 +1,60 @@
+import type { LexicalNode, NodeKey, SerializedTextNode } from 'lexical';
+import { TextNode } from 'lexical';
+
+export class PreviewMentionNode extends TextNode {
+  static getType(): string {
+    return 'previewmention';
+  }
```
shanpriyan, line 7: Can we?
```suggestion
  static getType(): 'previewmention' {
    return 'previewmention';
  }
```

src/components/Shared/Markdown/PreviewMentionNode.tsx
```diff
@@ -0,0 +1,60 @@
...
+  exportJSON(): SerializedTextNode {
+    return {
+      ...super.exportJSON(),
+      type: 'hashtag'
+    };
+  }
+
+  canInsertTextBefore(): boolean {
+    return false;
+  }
```
shanpriyan, line 47:
```suggestion
  canInsertTextBefore(): false {
    return false;
  }
```

src/components/Shared/Markdown/PreviewMentionPlugin.tsx
```diff
@@ -0,0 +1,256 @@
...
+    hashtagAlpha +
+    hashtagAlphanumeric +
+    '*)';
+
+  return hashtag;
+}
+
+const REGEX = new RegExp(getHashtagRegexString(), 'i');
+
+export function PreviewMentionPlugin(): JSX.Element | null {
```
shanpriyan, line 224:
```suggestion
export function PreviewMentionPlugin(): null {
```

bigint:
We have migrated to monorepo, can you please sync with `main` branch 🙏🏼

bigint:
sed laip @harish-sethuraman 😭

bigint:
@harish-sethuraman gonna close this for now 🙏🏼 

We can recreate new one on ref with this PR 🙇🏼

PR #1199, opened 2022-11-14
feat: add ability to mention user from his profile page

svitekpavel:
## What does this PR do?

Inspiration: Twitter

When you open user's profile, you are able to tweet him directly from his page. When you open "compose message", user's handle is pre-filled.

List of changes

- Add `NewPost` component to `pages/profile`
  - The component is not displayed when user is at his own profile
- Pre-fill `Editor` content to contain handle of user profile
- Set cursor to the right place

## Type of change

## How should this be tested?

- Test A: Create new message from default/Timeline page
- Test B: Create new message from user's profile page

## Checklist

✅ 

## Screenshots

Screenshot 2022-11-14 at 15 52 15

Screenshot 2022-11-14 at 15 52 29


bigint:
@harish-sethuraman can you check the lexical part 🙏🏼

src/components/Composer/Editor/index.tsx
```diff
@@ -47,4 +89,9 @@ const Editor: FC = () => {
   );
 };
 
+Editor.defaultProps = {
+  placeholder: '',
+  defaultContent: 'Hello world'
```
harish-sethuraman, line 94: Shouldn't this also be `Hello world` check the comment editor to see what happens 😛

src/components/Profile/index.tsx
```diff
@@ -76,6 +77,12 @@ const ViewProfile: NextPage = () => {
           <Details profile={profile as any} />
         </GridItemFour>
         <GridItemEight className="space-y-5">
+          {currentProfile && profile && currentProfile.id !== profile.id && (
```
harish-sethuraman, line 80: Why should we verify `currentProfile && profile` ? 
```suggestion
          {currentProfile?.id !== profile?.id && (
```
can we do it this way?

src/components/Profile/index.tsx
```diff
@@ -76,6 +77,12 @@ const ViewProfile: NextPage = () => {
...
+            <NewPost
+              defaultContent={`@${profile?.handle} `}
+              placeholder={`Write a message to @${profile?.handle}`}
```
harish-sethuraman, line 83: Im wondering what placeholder we should give here. Ideally we are not sending a message to the user rather we are mentioning the user in a post. Should we change it to something else @bigint?

src/components/Composer/Editor/index.tsx
svitekpavel: Right, forgot to remove this string. I set it to empty string.

src/components/Profile/index.tsx
svitekpavel: @harish-sethuraman if you keep only statement `currentProfile?.id !== profile?.id`: if `currentProfile == undefined` and `profile == undefined` then this statement will become `true` which is the case I was trying to avoid.

But I see there's `if (!data?.profile) { 404 }` so we can assume that at least `profile` is not empty, so we can change the statement like you suggested.

src/components/Profile/index.tsx
svitekpavel: Maybe `Mention @handle in a post` ?

harish-sethuraman: We have a little problem going on here. You can see that the mention is not actually a mentionNode and suggestions are popping up. May be we should make use of `createMentionNode` and create a mention node. Right now it is a paragraph node?

svitekpavel: Instead of pulling the logic all the way up here, can't we rather "parse" the default string in `Editor` component and convert the "@mention" string to a `MentionNode` ? It would create a more usable dev experience.

harish-sethuraman: We can do that. If we parse the `@mention` to be `MentionNode` we right now won't know if it actually is a valid mention or not while constructing in the editor. random text with `@asdasd` will also be considered as `MentionNode`. Rather we want to keep only valid mentions while creating a post. That is why we have a type ahead menu with suggestions and then convert it to `MentionNode` making sure it is a valid mention while entering it into editor.

svitekpavel: Right. I will have a look later today if I can create the `MentionNode` directly and update the prop types.

Another way I started thinking about - to simplify the value propagation to `Editor` - isn't there a way to set `EditorState` somehow in lenster custom state?

svitekpavel: @harish-sethuraman I tried to implement `$createMentionNode` directly in `Profile` component, but there's a problem that `Editor` doens't exists yet.

So I implemented 2nd solution - which parses the `@mention` and creates `TextNode` and `MentionNode` according to the parsed `defaultContent` string. The solution works properly (parsing), but as you mentioned above, it will create `MentionNode` even for strings which are not a correct profile. I'm not sure if that has any side effects for posting messages.

**Screenshot of the problem with missing `Editor`**

Screenshot 2022-11-15 at 16 34 33

svitekpavel: @harish-sethuraman shall we merge this?

svitekpavel: @harish-sethuraman thinking about the parsing topic. If/when we have case where you would like to avoid parsing of `@mention` to `MentionnNode` we can implement prop `disableParsing` or find other way how to go around this. However, until we have such a case, I think we can merge this in.

harish-sethuraman: Give me some time. I'll look into this in the weekend.

bigint:
We have migrated to monorepo, can you please sync with `main` branch 🙏🏼

svitekpavel:
> feature/allow-to-create-message-from-user-profile-page

It's synced.

src/components/Profile/index.tsx
bigint: @harish-sethuraman are you gonna look @ this 🙇🏼?

harish-sethuraman: Have to 😭  not getting time

bigint:
We don't have any plans to ship this immediately 😢 

We will have global post button like twitter does and it will takes the context if they are in user page, else it will be a normal post modal.

Closing this now 🙏🏼

svitekpavel:
> We don't have any plans to ship this immediately 😢
> 
> We will have global post button like twitter does and it will takes the context if they are in user page, else it will be a normal post modal.
> 
> Closing this now 🙏🏼

Well, this is very disappointing @bigint. You are not merging a functionality which people are missing every day, however, as you said your replacement solution is not implemented as of today. What if instead of closing this, you merge this and once you have the replacement solution ready, we simply remove this version of the feature and simply move onto the next. 

**For broader consideration**
This is the 2nd PR from me without merging, even if both PRs were opened as I started the work, so you could have signalled early that the functionality will not be merged. Rather, you chose to let me believe that this PR will be merged; sync the branch several times with the changes on `main` branch, only to close it today. 

It would be valuable for anyones future potential contribution if you did a better communication and improve the project management. Right now I feel that both of our time was wasted with zero shipped value.

bigint:
> You are not merging a functionality which people are missing every day

We didn't receive any request for this in our [canny](https://feedback.lenster.xyz/) or from anywhere

> you merge this and once you have the replacement solution ready

Feature back and forth will lead to bad UX

> This is the 2nd PR from me without merging, even if both PRs were opened as I started the work

Always make sure to create issues and discuss them there, if all is good to go we can go with PR

Issue #341, opened 2022-08-09
Add next-i18next support


bigint:

foolo:
@bigint Is this issue something that's currently relevant? I have done some work on this issue locally and have a working concept, which I could share as a draft PR, if it's of any interest at the moment, that is.

PR #1216, opened 2022-11-15
fix: spellcheck

### LINKED ISSUE #278: Do typo check across the repo

foolo:
## What does this PR do?

Fixes #278
I ran CSpell in the whole repo, and fixed errors that were found.

I divided the fixes into two commits: One for UI strings, and one for code.
@bigint If you think that only the UI strings should be changed, and not the code, just tell and I can just remove the second commit from the PR.

Note: I have not proofread all the files manually :) I think that a manual proofread can be done more easily after all UI strings have been moved to i18next translation files (i.e. after issue #341 has been implemented)

Finally, there is one typo, "Pubication",  in a GraphQL call in src/components/Shared/Pending.tsx:
```
const HAS_PUBLICATION_INDEXED_QUERY = gql`
  query HasPubicationIndexed($request: PublicationQueryRequest!) {
```
But it seems like fixing this typo might break the query, so I left it unchanged.

## Type of change


bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #1218, opened 2022-11-16
feat: Modified Shop Collects button to only show on minted out publications

ahomentc:
## What does this PR do?

Makes the Shop Collects button only appear for publications that satisfy these conditions:
1. The collects are all sold out
2. For timed collects, the time to collect has passed

This makes the Shop Collects button appear only when it is relevant, which is when users can't collect.
If users can't collect but still want the NFT, it's useful for them to have a handy link that takes them to the secondary marketplace.

## Type of change

## How should this be tested?
Open a publication who's collects have all been minted or a publication who's collect time has ended.


bigint:
Right now we don't have plans for this, we have some great ideas for future 🙇🏼 

Closing this now 🙏🏼

PR #1219, opened 2022-11-16
chore: hide .lens from username

bigint:


src/components/Shared/Slug.tsx
```diff
@@ -16,7 +16,7 @@ const Slug: FC<Props> = ({ slug, prefix, className = '' }) => {
       )}
     >
       {prefix}
-      {slug}
+      {slug.replace('.lens', '')}
```
harish-sethuraman, line 19: But why? 🤦🏻

bigint: Technically??

If it's a product-related question, yes, it's a request from the Lens product team 🙇🏼

bigint:
Closing now as a part of cleanup will open when team gets back 🚀

PR #1221, opened 2022-11-16
feat: add virtual list for timelines

bigint:


src/components/Explore/Feed.tsx
```diff
@@ -70,13 +67,23 @@ const Feed: FC<Props> = ({ focus, feedType = PublicationSortCriteria.CuratedProf
...
-          />
-        ))}
+      <Card>
+        <Virtuoso
+          useWindowScroll
+          className="virtual-list"
+          totalCount={publications?.length}
+          components={{ Footer: () => <PublicationsShimmer inVirtualList /> }}
+          itemContent={(index) => {
+            const publication = publications?.[index] as LensterPublication;
```
harish-sethuraman, line 77: Why are we type casting again in line number 82?

src/components/Home/Timeline/index.tsx
```diff
@@ -91,13 +88,23 @@ const Timeline: FC = () => {
...
-            publication={publication.root as LensterPublication}
-          />
-        ))}
+        <Virtuoso
+          useWindowScroll
+          className="virtual-list"
+          totalCount={publications?.length}
+          components={{ Footer: () => <PublicationsShimmer inVirtualList /> }}
+          itemContent={(index) => {
+            const publication = publications?.[index] as FeedItem;
```
harish-sethuraman, line 97: Typecasting twice here as well

src/components/Publication/SinglePublication.tsx
```diff
@@ -50,7 +53,12 @@ const SinglePublication: FC<Props> = ({
     : publication?.createdAt;
 
   return (
-    <article className="hover:bg-gray-100 dark:hover:bg-gray-800 cursor-pointer first:rounded-t-xl last:rounded-b-xl p-5">
+    <article
+      className={clsx(
+        { 'first:rounded-t-xl': index === 0 },
```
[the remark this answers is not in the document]
harish-sethuraman: Whilst you have already given `first:` in css why are we trying to use index here? Cant we do it in this component's parent to apply for all first element?

src/styles.css
```diff
@@ -175,3 +175,11 @@ i.code {
...
   @apply outline-none;
 }
+
+.virtual-list {
+  @apply overflow-y-hidden;
+}
+
+.virtual-list > div > div {
+  @apply divide-y-[1px] dark:divide-gray-700/80;
+}
```
harish-sethuraman, line 185: We have the same class in `Card` component as well? Do we need it at both places?

src/components/Explore/Feed.tsx
bigint: My wrong fixed 😊

src/components/Publication/SinglePublication.tsx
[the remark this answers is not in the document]
bigint: we can't use `first:` in parent or else we can use child selector but we need to do 7 level dom drill which is not same in all cases, better index is good for now 🙇🏼

src/styles.css
bigint: We are not using card inside virtual list and single publication, so we need this 🙇🏼

PR #1225, opened 2022-11-16
fix: code scanner alert

foolo:
## What does this PR do?

Fixes one of the alerts from the code scanner:
Screenshot_2022-11-16_19-19-09

@bigint I also looked at the other alert (for `src/components/Shared/Login/New/Pending.tsx:33`) but I'm not sure what the best solution for that one is. In the meanwhile, since the alerts could be security risks, this PR solves at least one of them.

## Type of change


bigint:
I'll fix this today 🙏🏼 

Closing this 🙇🏼

PR #1242, opened 2022-11-18
fix: crash on some posts containing markdown code formatting

### LINKED ISSUE #1236: can't access my profile any more

### Current Behavior

the app crashes

### Expected Behavior

it doesn't crash :)

### Steps To Reproduce

access https://lenster.xyz/u/sakulstra.lens

Web

Chrome

### Anything else?

image

foolo:
## What does this PR do?

Fixes #1236
The problem was that MDCodeMatcher was not marked as "void" (see https://interweave.dev/docs/matchers ) so it allowed sub-tags, which caused the crash. Also, the MDCodeMatcher need to be executed first, since a code snippet can contain any syntax, that could otherwise be caught by other matchers.

## Type of change

## How should this be tested?
Visit https://lenster.xyz/posts/0x101e-0x07
Confirm that the page doesn't crash, but shows some code snippets with asterisk inside:
Screenshot_2022-11-18_13-06-17

### DIFF
--- a/apps/web/src/components/utils/matchers/markdown/MDCodeMatcher.tsx
+++ b/apps/web/src/components/utils/matchers/markdown/MDCodeMatcher.tsx
@@ -13,8 +13,13 @@ export class MDCodeMatcher extends Matcher {
   }
 
   match(value: string) {
-    return this.doMatch(value, /`(.*?)`/u, (matches) => ({
-      match: matches[1]
-    }));
+    return this.doMatch(
+      value,
+      /`(.*?)`/u,
+      (matches) => ({
+        match: matches[1]
+      }),
+      true
+    );
   }
 }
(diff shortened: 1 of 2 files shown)


apps/web/src/components/utils/matchers/markdown/MDCodeMatcher.tsx
```diff
@@ -13,8 +13,13 @@ export class MDCodeMatcher extends Matcher {
...
-      match: matches[1]
-    }));
+    return this.doMatch(
+      value,
+      /`(.*?)`/u,
+      (matches) => ({
+        match: matches[1]
+      }),
+      true
+    );
```
bigint, line 23: Any changes here? 🤔  I'm not seeing any 😕

foolo: Might be hard to spot :) but I added the "true" at the end (which marks the formatter as "void").
And because of the added parameter, the Prettier formatter required all the extra line breaks.

PR #1264, opened 2022-11-21
feat: add prominent ENS badge to user profile and timeline

svitekpavel:
## What does this PR do?

**Community feature request**

- https://lenster.xyz/posts/0x017cc9-0x6a

If ENS name matches the lens handle, display ENS icon in Timeline and in Profile, next to "verified" badge.

## Type of change

## How should this be tested?

- Test A: Open Vitalik's profile
- Test B: Open Vitalik's post in the feed

vitalik-profile
vitalik-timeline-feed

### DIFF
--- a/apps/web/src/lib/isEnsVerified.ts
+++ b/apps/web/src/lib/isEnsVerified.ts
@@ -1,0 +1,20 @@
+import type { Profile } from 'lens';
+
+/**
+ *
+ * @param profile - Profile id
+ * @returns `true` if handle is same as ENS name
+ *
+ * Examples for handle "vitalik.lens":
+ *  - ENS: "vitalik.eth" returns `true`
+ *  - ENS: "lens.vitalik.eth" returns `true`
+ *  - ENS "vitalik.eth.somebodyelse.eth" returns `false` (trying to impersonate)
+ */
+function isEnsVerified(profile: Profile): boolean {
+  const ensName = (profile?.onChainIdentity?.ens?.name || '') as string;
+  const handleWithEnsSuffix = profile?.handle.replace(/\.lens$/, '.eth');
+
+  return ensName.endsWith(handleWithEnsSuffix);
+}
+
+export default isEnsVerified;
--- a/apps/web/src/components/Profile/Details.tsx
+++ b/apps/web/src/components/Profile/Details.tsx
@@ -1,4 +1,5 @@
 import Message from '@components/Profile/Message';
+import EnsCheckIcon from '@components/Shared/EnsCheckIcon';
 import Follow from '@components/Shared/Follow';
 import Markup from '@components/Shared/Markup';
 import Slug from '@components/Shared/Slug';
@@ -16,6 +17,7 @@ import { buildConversationKey } from '@lib/conversationKey';
 import formatAddress from '@lib/formatAddress';
 import getAttribute from '@lib/getAttribute';
 import getAvatar from '@lib/getAvatar';
+import isEnsVerified from '@lib/isEnsVerified';
 import isStaff from '@lib/isStaff';
 import isVerified from '@lib/isVerified';
 import { STATIC_IMAGES_URL } from 'data/constants';
@@ -82,6 +84,11 @@ const Details: FC<Props> = ({ profile }) => {
           {isVerified(profile?.id) && (
             <Tooltip content="Verified">
               <BadgeCheckIcon className="w-6 h-6 text-brand" />
+            </Tooltip>
+          )}
+          {isEnsVerified(profile) && (
+            <Tooltip content={`ENS: ${profile?.onChainIdentity?.ens?.name}`}>
+              <EnsCheckIcon className="w-6 h-6" alt={`ENS: ${profile?.onChainIdentity?.ens?.name}`} />
             </Tooltip>
           )}
         </div>
--- a/apps/web/src/components/Shared/EnsCheckIcon.tsx
+++ b/apps/web/src/components/Shared/EnsCheckIcon.tsx
@@ -1,0 +1,18 @@
+import { STATIC_IMAGES_URL } from 'data/constants';
+import * as React from 'react';
+
+interface Props {
+  alt?: string;
+  className?: string;
+}
+
+const EnsCheckIcon: React.FC<Props> = ({ className, alt }) => {
+  return <img src={`${STATIC_IMAGES_URL}/badges/ens.png`} className={className} alt={alt} />;
+};
+
+EnsCheckIcon.defaultProps = {
+  alt: '',
+  className: ''
+};
+
+export default EnsCheckIcon;
--- a/apps/web/src/components/Shared/UserProfile.tsx
+++ b/apps/web/src/components/Shared/UserProfile.tsx
@@ -1,6 +1,7 @@
 import { BadgeCheckIcon } from '@heroicons/react/solid';
 import getAttribute from '@lib/getAttribute';
 import getAvatar from '@lib/getAvatar';
+import isEnsVerified from '@lib/isEnsVerified';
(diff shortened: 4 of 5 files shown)


apps/web/src/lib/isEnsVerified.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @param profile - Profile id
+ * @returns `true` if handle is same as ENS name
+ *
+ * Examples for handle "vitalik.lens":
+ *  - ENS: "vitalik.eth" returns `true`
+ *  - ENS: "lens.vitalik.eth" returns `true`
+ *  - ENS "anothervitalik.eth" returns `false` (trying to impersonate)
+ *  - ENS "vitalik.eth.somebodyelse.eth" returns `false` (trying to impersonate)
+ */
+function isEnsVerified(profile: Profile): boolean {
```
paeet, line 14: Hi Pavel, I would like to open the discussion on the condition of this function.

In my opinion, the ENS verified batch should not be dependent on whether ENS and lens handles do match. 

You would be excluded from the functionality if:
- someone squats your ENS handle on lens
- you loose/change your lens handle

It should show up for every user, whose address resolves to an ENS name.

wdyt?

paeet:
Hi Pavel, 

first of all thank you for taking the initiative on this one!

I think it would make sense to add some sort of **priority** to the various level of account verifications. Based on this priority, I would only show one level of account verification at a time.

**Priority:**
- 1 -> purple checkmark
- 2 -> ENS account name

**For Example:**
- vitalik.lens both has **purple checkmark** and an **ENS account name**. Only **purple checkmark** will be shown next to account name
- 59401.lens only has an **ENS account name**. **ENS account name** is shown.

wdyt?

BR,
fausto

apps/web/src/lib/isEnsVerified.ts
svitekpavel: Hi @faustofren .. thanks for the feedback

1. To be able to identify if somebody is squating your ENS handle, it would mean that we would need to do additional "search" every time we are creating the ENS/lens identifiers. For example: If you open one user profile, we would need to look for other people based on your LENS handle. Even worse, when displaying a Timeline feed, we would need to do a lot of these searches which would create unnecessary load for APIs. Of course, there would be a possibility to create our own database where we would store something like `isThisUserSquatting: bool` but that would require creating such a database and it would defeat a bit the purpose of having all these data available in native Lens protocol.

2. We should not treat the "ENS name" somehow superior, above everything else. If you own some ENS name, it doesn't necessarily mean that you are entitled to such handle / string in every single naming system / social media. It would rather defeat the purpose of decentralization. But, of course, because of it's history, adoption, popularity; many people think of ENS just like that.

svitekpavel:
> Hi Pavel,
> 
> first of all thank you for taking the initiative on this one!
> 
> I think it would make sense to add some sort of **priority** to the various level of account verifications. Based on this priority, I would only show one level of account verification at a time.
> 
> **Priority:**
> 
> * 1 -> purple checkmark
> * 2 -> ENS account name
> 
> **For Example:**
> 
> * vitalik.lens both has **purple checkmark** and an **ENS name**. Only **purple checkmark** will be shown next to account name
> * 59401.lens only has an **ENS name**. **ENS name** is shown.
> 
> wdyt?
> 
> BR, fausto

I like your way of thinking. I also started thinking about a concern where we would like to display more of these badges, suddenly resulting in having 6 badges next to each other - for better or worse; which would also create UX/UI challenges. I would agree to find some reasonable tactic of prioritisation what to display.

apps/web/src/lib/isEnsVerified.ts
faustofren: Hi Pavel,

1. Agree, don´t think any specific application side is necessary here.

2. Agree, it´s not superior above everything else. We just have to view each level of verification as it´s own and prioritize them.

The ENS name indeed actually verifies your **pseudonymous** identity, including your Ethereum on-chain activity (DAO participation, NFT collecting, Defi participation etc.) up to the past 5 years.

Vitalik could have actually chosen a completely random lens handle. The ENS name verification still would have ensured that this account actually is controlled by Vitalik himself, due to the provenance and significance of his ENS name. No other verification would have been required actually.

The ENS name badge could also be helpful for crypto/web3 sub-communities (DAOs, projects etc.) who are mostly operating pseudonymously to discover and verify each other, without the need to hold a 1:1 ENS/lens handle copy.

wdyt?

BR,
fausto

faustofren:
> > Hi Pavel,
> > first of all thank you for taking the initiative on this one!
> > I think it would make sense to add some sort of **priority** to the various level of account verifications. Based on this priority, I would only show one level of account verification at a time.
> > **Priority:**
> > 
> > * 1 -> purple checkmark
> > * 2 -> ENS account name
> > 
> > **For Example:**
> > 
> > * vitalik.lens both has **purple checkmark** and an **ENS name**. Only **purple checkmark** will be shown next to account name
> > * 59401.lens only has an **ENS name**. **ENS name** is shown.
> > 
> > wdyt?
> > BR, fausto
> 
> I like your way of thinking. I also started thinking about a concern where we would like to display more of these badges, suddenly resulting in having 6 badges next to each other - for better or worse; which would also create UX/UI challenges. I would agree to find some reasonable tactic of prioritisation what to display.

agreed, ultimately, and I am not sure if possible, it would be great if you could choose the level of verification (badge) yourself in the lenster settings.

would this be possible?

apps/web/src/lib/isEnsVerified.ts
svitekpavel: @faustofren I agree with both points and verified ENS badge is in places like `Timeline` and `Profile` should definitely help people from communities identify each other easier. I think we should display the ENS name in `Timeline` (currently it's not). It's already displayed in `Profile` page.

svitekpavel:
> > > Hi Pavel,
> > > first of all thank you for taking the initiative on this one!
> > > I think it would make sense to add some sort of **priority** to the various level of account verifications. Based on this priority, I would only show one level of account verification at a time.
> > > **Priority:**
> > > 
> > > * 1 -> purple checkmark
> > > * 2 -> ENS account name
> > > 
> > > **For Example:**
> > > 
> > > * vitalik.lens both has **purple checkmark** and an **ENS name**. Only **purple checkmark** will be shown next to account name
> > > * 59401.lens only has an **ENS name**. **ENS name** is shown.
> > > 
> > > wdyt?
> > > BR, fausto
> > 
> > 
> > I like your way of thinking. I also started thinking about a concern where we would like to display more of these badges, suddenly resulting in having 6 badges next to each other - for better or worse; which would also create UX/UI challenges. I would agree to find some reasonable tactic of prioritisation what to display.
> 
> agreed, ultimately, and I am not sure if possible, it would be great if you could choose the level of verification (badge) yourself in the lenster settings.
> 
> would this be possible?

I was thinking the same way - let user decide which verification method(s)/badges are most relevant to him/her. I think that this should be technically possible.

apps/web/src/lib/isEnsVerified.ts
faustofren: also to add, matching ens name and lens handle is logically not viable because of the different length requirements of the two handles.

- ENS: minimum 3 characters
- lens: minimum 5 characters

bigint:
We want to keep the lenster UI simple and accessible, if we add more badges it will not be a good idea 🙏🏼

bigint:
@faustofren we will implement [Cultivator DAO](https://mirror.xyz/lensprotocol.eth/pIzwjs9uhH8eSS1Bx0K1r-3iHEgLlBt6ruPXcEJe0S0) soon, so it will become permission less

If we display ENS after some time users will request to show Proof of humanity and after that Sybil badge, which makes the UI worse 🙇🏼

PR #1358, opened 2022-12-05
fix: lazy load dm convo list

bhavya2611:
## What does this PR do?

Adds lazy loading to fetch the preview messages for the conversations list

## Type of change

## How should this be tested?

- [ ] See if everything on messaging works correctly
- [ ] If you have a lot of conversations see if the loading is quicker and smoother


apps/web/src/components/utils/hooks/useMessagePreviews.tsx
```diff
@@ -3,40 +3,37 @@ import buildConversationId from '@lib/buildConversationId';
...
   const messageProfiles = useMessageStore((state) => state.messageProfiles);
   const setMessageProfiles = useMessageStore((state) => state.setMessageProfiles);
   const previewMessages = useMessageStore((state) => state.previewMessages);
   const setPreviewMessages = useMessageStore((state) => state.setPreviewMessages);
   const selectedProfileId = useMessageStore((state) => state.selectedProfileId);
-  const setSelectedProfileId = useMessageStore((state) => state.setSelectedProfileId);
   const setPreviewMessage = useMessageStore((state) => state.setPreviewMessage);
-  const reset = useMessageStore((state) => state.reset);
-  const { client, loading: creatingXmtpClient } = useXmtpClient();
+  const { client } = useXmtpClient();
```
elisealix22, line 26: We only want to call `useXmtpClient()` once to avoid multiple signature reqs from multiple hooks. Since the loading state is now managed in `useListConversations`, this can be:

```suggestion
  const { client } = useMessageStore((state) => state.client);
```

apps/web/src/components/utils/hooks/useStreamConversations.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { useAppStore } from 'src/store/app';
+import { useMessageStore } from 'src/store/message';
+
+import useXmtpClient from './useXmtpClient';
+
+const useStreamConversations = () => {
+  const currentProfile = useAppStore((state) => state.currentProfile);
+  const conversations = useMessageStore((state) => state.conversations);
+  const setConversations = useMessageStore((state) => state.setConversations);
+  const { client } = useXmtpClient();
```
elisealix22, line 14: Same here:

```suggestion
  const client = useMessageStore((state) => state.client);
```

apps/web/src/components/Messages/PreviewList.tsx
```diff
@@ -34,10 +35,36 @@ const PreviewList: FC<Props> = ({ className, selectedConversationKey }) => {
...
+    (node) => {
+      if (previewLoading) {
+        return;
+      }
+      if (observer.current) {
+        observer.current.disconnect();
+      }
+      observer.current = new IntersectionObserver((entries) => {
+        if (entries[0].isIntersecting && hasMore) {
+          setCurrentIndex(currentIndex + 20);
```
elisealix22, line 58: Might be worth pulling this out in a constant since it's used in 2 different files.

elisealix22 (commented):
Noticed a couple of duplicate sig requests that can be resolved.

Also still investigating, but I created a profile on testnet that has 50+ convos and it's taking about ~10s to load. I'll investigate further with more logs, but here's the video. Additionally, it looks like the pagination isn't quite working on my branch (I'm capped at 20 convos).

https://user-images.githubusercontent.com/556051/205744081-26a71ebc-0965-4401-a136-3f6971e32d65.mov

elisealix22:
Getting closer: It looks like it's the `useListConversations` hook that is the bottleneck right now. Going to investigate how many requests we're still making in there 👀 

Screenshot 2022-12-05 at 5 21 12 PM

elisealix22:
Closing this out for now as we continue to investigate! For more context on where we're at, see this comment: https://github.com/lensterxyz/lenster/pull/1366#issuecomment-1342726519

PR #1366, opened 2022-12-06
fix: speed up conversation list

elisealix22:
## What does this PR do?
This PR is a WIP to debug conversation list latency. I was able to reproduce 10s + loading times with a wallet that has 51 conversations in it.

| Before (main) | After (ea/fix-conversations-lock) |
|--|--|
| Screenshot 2022-12-06 at 7 26 33 PM | Screenshot 2022-12-06 at 6 55 12 PM |

## Type of change

## How should this be tested?

- Use the preview link to load `/messages` page and ensure it loads faster than on lenster.xyz (especially @bigint who is very popular!)


neekolas:
I don't understand the part about 
> all awaiting on the same conversations object causing it to lock up

We don't have any explicit mutexes or locks on any objects. Are we sure that it wasn't just doing duplicate HTTP requests or something? The browser will throttle at some number of concurrent http reqs

apps/web/src/components/utils/hooks/useMessagePreviews.tsx
```diff
@@ -198,9 +198,13 @@ const useMessagePreviews = () => {
       }
     };
 
-    listConversations();
-    streamConversations();
-    streamAllMessages();
+    const setupConversationsList = async () => {
+      await listConversations();
+      await streamConversations();
```
neekolas, line 203: Wait. Neither `streamConversations` or `streamAllMessages` ever return. This will stop at line 203 and never make it to line 204

neekolas: You should be able to confirm by seeing if new messages are being streamed in for anything other than the currently open conversation.

elisealix22: Great catch. It should be the `list()` method that was causing trouble anyways. I'll reconfirm and add timing benchmarks to the description. It still should be significantly faster as-is still (just broken streaming).

elisealix22:
Updated to a WIP / draft PR but would still love to verify that it's significantly faster for Yoginth if possible! It seems specific to `streamAllMessages()` at this point.

bigint:
@elisealix22 I have the same 😕 

kinda same speed and more requests too

Screenshot 2022-12-08 at 11 52 09 AM

elisealix22:
Thank you so much for testing, Yoginth. Yeah, this wouldn't have removed any `query`s, only two `subscribe`s. Going to spend all of today on this today and then regroup with the team to brainstorm other solutions. So far we've ruled out:

1. Only loading 10 preview messages at a time doesn't resolve the issue -- the bottleneck appears to be in the listing of conversations.
2. It's not a concurrency problem (which is what I had originally thought here). Neither sequencing the `conversations.list()` or removing the duplicate calls speeds up Yoginth's time to see the conversation list.

elisealix22:
Closing for now as we continue to investigate!

PR #1367, opened 2022-12-07
fix: profile post duplication

### LINKED ISSUE #1031: Profile Posts Duplication

### Current Behavior

I am scrolling through my posts and they keep repeating in the same order as I have posted 

### Expected Behavior

Should show only once 

### Steps To Reproduce

Scroll your profile page and you would see all the posts duplication at least once 

Web

Chrome

### Anything else?

https://user-images.githubusercontent.com/100476953/198967172-b91c2865-ee7f-469e-b888-ea01f4069519.mp4

saurabhdoteth:
## What does this PR do?
returns profile publications uniquely by `id` preventing duplication

## Fixes
fixes #1031 

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B

## Checklist

- I haven't read the [contributing guide](https://github.com/lensterxyz/lenster/blob/main/CONTRIBUTING.md)
- My code doesn't follow the style guidelines of this project
- I haven't performed a self-review of my own code and corrected any misspellings
- I haven't commented on my code, particularly in hard-to-understand areas
- I haven't checked if my PR needs changes to the documentation
- I haven't checked if my changes generate no new warnings
- I haven't checked if new and existing unit tests pass locally with my changes

### DIFF
--- a/apps/web/src/components/Profile/Feed.tsx
+++ b/apps/web/src/components/Profile/Feed.tsx
@@ -59,7 +59,9 @@ const Feed: FC<Props> = ({ profile, type }) => {
     skip: !profile?.id
   });
 
-  const publications = data?.publications?.items;
+  const publications = [
+    ...new Map(data?.publications?.items?.map((publication) => [publication['id'], publication])).values()
+  ];
   const pageInfo = data?.publications?.pageInfo;
   const hasMore = pageInfo?.next && publications?.length !== pageInfo.totalCount;
 
--- a/packages/tsconfig/base.json
+++ b/packages/tsconfig/base.json
@@ -14,7 +14,8 @@
     "noUnusedParameters": false,
     "preserveWatchOutput": true,
     "skipLibCheck": true,
-    "strict": true
+    "strict": true,
+    "downlevelIteration": true
   },
   "exclude": ["node_modules"]
 }


apps/web/src/components/Profile/Feed.tsx
```diff
@@ -59,7 +59,9 @@ const Feed: FC<Props> = ({ profile, type }) => {
     skip: !profile?.id
   });
 
-  const publications = data?.publications?.items;
+  const publications = [
+    ...new Map(data?.publications?.items?.map((publication) => [publication['id'], publication])).values()
```
harish-sethuraman, line 63: Im guessing this should be validated from backend and not on frontend. Backend is not supposed to send same post twice? 

PS: If you were going to remove duplicates why not use `set` ?

saurabhdoteth: Yes makes sense to use ‘Set()’

bigint: But have we validated backend is sending twice? I think some problem with apollo cache 🙇🏼

bigint: @saurabhdoteth did you validate the root cause of the issue?

saurabhdoteth: yes, it seems like apollo is causing the issue because when I checked the network tab it returned the correct number of posts(without duplications) but when I console log'd `data` from `useProfileFeedQuery()` it contained twice the number of publications.

bigint: This fixed in https://github.com/lensterxyz/lenster/pull/1395

bigint:
This fixed in https://github.com/lensterxyz/lenster/pull/1395

@harish-sethuraman @saurabhdoteth 🙇🏼

PR #1379, opened 2022-12-08
fix: support attaching same images

ercwangwh:
## What does this PR do?
I can't attach the same image to my publication, like twitter.
I've tried to fix it and support attaching same image.
Please take a look if it is feasible.

Fixes # (issue)
image
image

## Type of change

## How should this be tested?

- [ ] Test A
- [ ] Test B


apps/web/src/components/Shared/Attachments.tsx
```diff
@@ -91,7 +91,7 @@ const Attachments: FC<Props> = ({
                 },
                 'relative'
               )}
-              key={url}
+              key={index}
```
harish-sethuraman, line 94: Do not use index as `key` please change it to url itself, If the url remains the same use `url + index`

ercwangwh: Thanks for the suggestion, I've updated

bigint: Thanks for suggestion @harish-sethuraman 🙏🏼 

Do yu have anything else?

apps/web/src/components/Composer/Actions/Attachment.tsx
```diff
@@ -64,23 +64,44 @@ const Attachment: FC<Props> = ({ attachments, setAttachments }) => {
     return false;
   };
 
+  const isImageType = (files: any) => {
+    for (const file of files) {
+      if (ALLOWED_IMAGE_TYPES.includes(file.type)) {
+        return true;
+      }
+    }
```
harish-sethuraman, line 72: When we select multiple files and `isImageType` is called with multiple `files` it will return true if first image file type is found 😅 is this expected? Because the false is returned only when none of the files are image. but true is returned if any one of the file type is image?

ercwangwh: Thank you for pointing out my mistake, I didn't consider the case where the first selected file is not an image when selecting multiple files. I will carefully check my code logic and test it before submitting it next time. 🙏

harish-sethuraman: Yea NP lets fix it rest looks good to me ✅

bigint:
@harish-sethuraman is it good to go?

apps/web/src/components/Composer/Actions/Attachment.tsx
```diff
@@ -64,23 +64,44 @@ const Attachment: FC<Props> = ({ attachments, setAttachments }) => {
...
     setLoading(true);
 
     try {
       // Count check
-      if (evt.target.files && (hasVideos(evt.target.files) || evt.target.files.length > 4)) {
-        return toast.error('Please choose either 1 video or up to 4 photos.');
+      if (
+        evt.target.files &&
+        (hasVideos(evt.target.files) ||
+          (isImageType(evt.target.files) && evt.target.files.length + attachments.length > 4))
```
harish-sethuraman, line 87: nit: assign `evt.target.files` to a const and use that

harish-sethuraman (approved):
LGTM

apps/web/src/components/Composer/Actions/Attachment.tsx
bigint: @ercwangwh can you do this 🙇🏼

ercwangwh: @bigint updated 🙏

bigint (approved):

PR #1386, opened 2022-12-09
fix: case insensitive checking 

harryranakl:
## What does this PR do?
lib: formatHandle

Fixes # (issue)
Screenshot 2022-12-09 at 2 56 42 PM

## Type of change

### DIFF
--- a/apps/web/src/lib/formatHandle.ts
+++ b/apps/web/src/lib/formatHandle.ts
@@ -11,7 +11,7 @@ const formatHandle = (handle: string | null, keepSuffix = false): string => {
     return '';
   }
 
-  if (handle === LENSPROTOCOL_HANDLE) {
+  if (handle?.toLowerCase() === LENSPROTOCOL_HANDLE?.toLowerCase()) {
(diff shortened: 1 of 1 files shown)


apps/web/src/lib/formatHandle.ts
```diff
@@ -11,7 +11,7 @@ const formatHandle = (handle: string | null, keepSuffix = false): string => {
     return '';
   }
 
-  if (handle === LENSPROTOCOL_HANDLE) {
+  if (handle?.toLowerCase() === LENSPROTOCOL_HANDLE?.toLowerCase()) {
```
bigint, line 14: We already know LENSPROTOCOL_HANDLE is lowercase, do we need to check this?

bigint: @harryranakl

bigint:
```suggestion
  if (handle?.toLowerCase() === LENSPROTOCOL_HANDLE) {
```

PR #1391, opened 2022-12-09
Update xmtp-js

neekolas:
## What does this PR do?

`xmtp-js` 7.4.0 contains some significant performance improvements for users with a large number of active conversations. Conversations are now cached, so that repeated calls to `client.conversations.list()` do not have to download or decrypt the same conversation multiple times.

It's not a complete solution to the problem, but it should cut down the CPU load by >50% for someone like @bigint.

Fixes # (issue)

## Type of change


bigint:
Amazing update 🚀

PR #1448, opened 2022-12-22
feat: update atob api

ylmin:
## What does this PR do?
This function is only provided for compatibility with legacy web platform APIs and should never be used in new code, because they use strings to represent binary data and predate the introduction of typed arrays in JavaScript. For code running using Node.js APIs, converting between base64-encoded strings and binary data should be performed using Buffer.from()

## How should this be tested?
Check that the warning is gone in code scanning results


bigint:
I see lot of changes 🤔 

can you fix the PR 🙏🏼

PR #1532, opened 2023-01-05
fix: show login modal on follow

### LINKED ISSUE #938: Prompt to login in alongside login error message.
### Summary

When a user attempts to perform an action and has not logged in, rather than show an error asking them to login, show the error and trigger the login process.

In this example the user is not logged in but tries to follow.

image

foolo:
## What does this PR do?

Fixes #938
When user is not logged in, and presses Follow or Super follow: Instead of showing "Please sign in your wallet.", open the login dialog to let the user sign in. The user will then need to press Follow again.
[Screen recording](https://user-images.githubusercontent.com/667227/210740703-3fe9c7cf-eaaf-405a-8471-c0787e1c2564.mp4)

## Type of change


bigint:
Amazing can we do the same for mirror and likes 🙏🏼 ?

foolo:
> Amazing can we do the same for mirror and likes 🙏🏼 ?

@bigint I actually tried it first but realized that the solution is not as straight-forward as for Follow, so I decided to not include it in this PR.
If the user is logged out, and clicks "Like" on a post on the start page (directly on https://lenster.xyz/), and we simply show the Login modal, then, when user has signed in, they will be taken to their Timeline, and then they will probably no longer see the same post as they clicked "Like" on on the logged-out start page. So two possible solutions that I can think of:
- Propagate the URL of the post that "Like" was clicked on, and take user to the post's page after logging in
- Simply disable Like, Mirror, etc, when logged out, and just show "5 likes", etc.

PR #1540, opened 2023-01-05
fix: persist tab in explore

### LINKED ISSUE #997: Add persisted tab via url query param
We need to persist the tab state via URL query param, example if user switch to trending tab in explore we need to persist that in URL query param so it will work even after page reload

foolo:
## What does this PR do?

Fixes #997
Save selected tab in http://localhost:4783/explore as a URL query parameter.
[Screen recording](https://user-images.githubusercontent.com/667227/210841401-fba1d09b-e59e-41c0-b8f2-df393b77b28b.mp4)

Note: Using `router.replace` so that the user can click around among the tabs without creating more browser history. If this is not the desired behavior, it can simply be changed to `router.push`.

## Type of change

### DIFF
--- a/apps/web/src/components/Explore/index.tsx
+++ b/apps/web/src/components/Explore/index.tsx
@@ -11,6 +11,7 @@ import clsx from 'clsx';
 import { APP_NAME, STATIC_IMAGES_URL } from 'data/constants';
 import { PublicationSortCriteria } from 'lens';
 import type { NextPage } from 'next';
+import { useRouter } from 'next/router';
 import { useState } from 'react';
 import { useAppStore } from 'src/store/app';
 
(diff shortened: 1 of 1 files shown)


foolo:
Also note that there is an alternative solution which may be easier to read:
```
router.query.tabIndex = index.toString();
router.replace(router);
```
but that creates a big amount of warnings "Unknown key passed via urlObject into url.format"
https://stackoverflow.com/a/66382571/961254

bigint:
Link T-2101

bigint:
> but that creates a big amount of warnings "Unknown key passed via urlObject into url.format"

Ahh this makes sense 🚀

bigint (approved):
**LGTM** :shipit: 

Rad LGTM

foolo:
> Link T-2101

@bigint  Just out of curiosity, what does this mean?  :)

PR #1563, opened 2023-01-07
fix: "You can't reply to this post" not being visible in dark mode

### LINKED ISSUE #1562: Can't see "you can't reply to this post" in dark mode

### Current Behavior

The text color of the "You can't reply to this post" warning is not visible in dark mode.

image
image

### Expected Behavior

Text color should be darker for the blueish background

### Steps To Reproduce

Open the post in dark mode

Web

Brave

### Anything else?

_No response_

kyrers:
## What does this PR do?

Hopefully, this will fix the `"You can't reply to this post"` text not being visible in dark mode according to your style guidelines and design choices.

I tested in light mode as well, and it is still visible with this change.

Fixes #1562 

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)
- [X ] Enhancement (non-breaking small changes to existing functionality)

## How should this be tested?

To make sure the issue exists, open a gated post in dark mode and verify that you can't see the  `"You can't reply to this post"` text. Then, do the same with this PR applied and you should now be able to see the text in a greyish color.


apps/web/src/components/Shared/CommentWarning.tsx
```diff
@@ -5,7 +5,7 @@ import type { FC } from 'react';
 
 const CommentWarning: FC = () => {
   return (
-    <Card className="!bg-blue-100 border-blue-300 flex items-center space-x-1.5 text-sm font-bold lt-text-gray-500 p-5">
+    <Card className="!bg-blue-100 border-blue-300 flex items-center space-x-1.5 text-sm font-bold text-gray-500 p-5">
```
harish-sethuraman, line 8: Actually `lt-text-gray-500` is also a valid class. We have to make sure if the same problem is happening at other places where we have used `lt-text-gray-500`? Would you like to look into that as well instead of removing it here alone?

kyrers: With `lt-text-gray-500` the `color` attribute gets overridden by the `.dark .lt-text-gray-500` class which, aside from making total sense, does not happen when using `text-gray-500`.

I fixed it this way to avoid causing problems in other UI components. 

I did look around for other possible places where `lt-text-gray-500` might not be visible such as the `CollectForm` and `Mirrored` components, but everything seems to be ok because the background is actually dark and not blue. I did not check every single place where `lt-text-gray-500` is used though.

kyrers: Ok, I decided to check every place where `lt-text-gray-500` is used and I found at least one more place where it doesn't work. The cause is, again, the background not being neither black or white.

I can't right now, but tomorrow I will check the usages I haven't checked yet and fix them before adding another PR.

Adding `!important` to `lt-text-gray-500`  would probably fix the issue, but I try to avoid `!important` whenever possible. So I will stick to replacing `lt-text-gray-500` with `text-gray-500` until told otherwise because I think this is the better practice.

harish-sethuraman: Thank you so much for your efforts 🎉 . Take your time on this and lets fix this at other places as well 😄

bigint:
Link T-2460

apps/web/src/components/Shared/CommentWarning.tsx
kyrers: I just pushed 65f6456, which fixes the same issue in the only other component I found it.

I looked through every usage of `lt-text-gray-500` including staff tools, and I'm confident this specific issue isn't happening anywhere else. 

Hopefully, this is enough to fix the issue. Let me know if I can help in any other way.

harish-sethuraman: Thank you so much for taking time on this 👍🏻

bigint: Thanks for fixing this @kyrers and thanks for hoping here @harish-sethuraman 

LGTM 🚀

bigint (approved):

PR #1570, opened 2023-01-08
fix: Fix typos

rex4539:


bigint:
This is issue on API, we use `yarn:codegen` to generate this file automatically from the backend, its not on lenster side 🙏🏼

PR #1600, opened 2023-01-12
fix: url matcher for parentheses

### LINKED ISSUE #1599: URL detection/rendering not working for some URLs

### Current Behavior

URLs with parentheses not detected correctly, and also links to the wrong destination.

Screenshot_2023-01-12_11-08-21
(https://lenster.xyz/posts/0x89dc-0x0517)

### Expected Behavior

Render URLs with parentheses correctly.
Should probably also be able to handle URLs inside parentheses. `(https://example.com)` or `(https://en.wikipedia.org/wiki/Dosa_(food))`

### Steps To Reproduce

Make a post with a URL with with parentheses, for example `https://en.wikipedia.org/wiki/Dosa_(food)` or `https://example.com/foo(bar)`

Web

Chrome

### Anything else?

_No response_

foolo:
## What does this PR do?

Fixes #1599

After fix:
Screenshot_2023-01-12_18-20-10
(purple color brightened for the screenshot to distinguish clearly from the black text)

## Type of change

## How should this be tested?

See steps in #1599

### DIFF
--- a/apps/web/src/components/Shared/Markup.tsx
+++ b/apps/web/src/components/Shared/Markup.tsx
@@ -6,7 +6,7 @@ import { MDLinkMatcher } from '@components/utils/matchers/markdown/MDLinkMatcher
 import { MDQuoteMatcher } from '@components/utils/matchers/markdown/MDQuoteMatcher';
 import { MDStrikeMatcher } from '@components/utils/matchers/markdown/MDStrikeMatcher';
 import { MentionMatcher } from '@components/utils/matchers/MentionMatcher';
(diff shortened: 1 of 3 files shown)


foolo:
Just realized this will incorrectly catch the last parentheses when the URL itself is inside parentheses `(https://example.com/foo)`.
Closing PR for now to come up with a better solution.

bigint:
Link T-3244

bigint (approved):

bigint:
LGTM 🚀

PR #1605, opened 2023-01-13
feat: add lock icon for gated publications

bigint:


bigint:
Close [T-2892](https://avara.height.app/T-2892)

PR #1606, opened 2023-01-13
fix: remove usage of LensterPublication type

bigint:


bigint:
Close [T-3481](https://avara.height.app/T-3481)

Issue #400, opened 2022-08-15
Site becomes dead after using coinbase wallet


bigint:

0xchetan:
The integration of #1340 should solve this issue.

bigint:
Hope so 🤞🏼

PR #1613, opened 2023-01-14
Rainbow wallet

### LINKED ISSUE #1340: Migrate to rainbow wallet
### Summary

Migrate to rainbow wallet

foolo:
## What does this PR do?

Add support for Rainbow Wallet and use RainbowKit's wallet selector
Fixes #1340 and #430 (and possibly also #400)

@bigint I'm opening a draft PR for this, since there are some open questions regarding the design.

Open questions:
- RainbowKit does not seem to support localization. With this PR however, only some parts of RainbowKit are used, so there are just a few strings visible from RainbowKit, ("Connect a Wallet", "Scan with Rainbow", etc.)
- RainbowKit can be configured for dark mode, but I have not found any way to switch between dark and light mode.
- Current PR design uses only the wallet selector (useConnectModal) from RainbowKit. We still use Lenster's current  Login button, and the account logic (handleSign, setCurrentProfile, etc) from WalletSelector (which should then probably be renamed :) ). In order to use the complete RainbowKit flow, it seems like we would need some callback methods, to plug in Lenster's account logic, but I have not found any such possibility.

## Type of change

## How should this be tested?

Go to Vercel deployment log in with RainbowWallet and other wallets.

## Checklist

- I haven't checked if my changes generate no new warnings


bigint:
@foolo can we automatically call the signing function once got connect to the wallet

You can refer in the Lenstube codebase as well https://github.com/lenstube-xyz/lenstube

cc @sasicodes

bigint:
Yea we can do the same as lenstube 🙇

foolo:
updated the behavior to start signing automatically
https://user-images.githubusercontent.com/667227/215323560-fadd5dc7-af58-4990-9d0c-2841cbe6dac7.mp4

apps/web/src/components/Shared/Follow.tsx
```diff
@@ -132,6 +134,7 @@ const Follow: FC<Props> = ({ profile, showText = false, setFollowing, followSour
 
   const createFollow = async () => {
     if (!currentProfile) {
+      openConnectModal?.();
```
foolo, line 137: @bigint  Now we need to call both **openConnectModal** and **setShowAuthModal**.
So we might want to move them to a custom hook.

```
openConnectModal?.();
setShowAuthModal(true);
```

apps/web/src/components/Shared/Login/WalletSelector.tsx
```diff
@@ -105,45 +80,32 @@ const WalletSelector: FC<Props> = ({ setHasConnected, setHasProfile }) => {
...
-            <Trans>Sign-In with Lens</Trans>
-          </Button>
+          <div className="text-sm font-bold text-center flex">
+            <div className="flex">
+              <Spinner size="sm" className="mx-auto mr-2" />
+              <Trans>Signing …</Trans>
+            </div>
+          </div>
         ) : (
           <SwitchNetwork />
```
foolo, line 107: SwitchNetwork may be removed too, if we are fine with letting the wallet take care of the "switch network" dialog (although note that some wallets, like Coinbase Wallet, switches network directly without prompting the user )

foolo:
@bigint This is good for review now! Now, signing is called automatically, and the flow requires less clicking.

bigint:
Sorry for the delay, gonna review this now 🙇🏼

bigint:
Link T-1549

bigint:
https://user-images.githubusercontent.com/69431456/218432331-541ebfb7-36ed-4576-ac5b-7e0dd42d8f01.mov



Disconnect wallet actually not working 🙇🏼

bigint:
image

Messed up a bit 🙇🏼

foolo:
> Screen.Recording.2023-02-13.at.3.50.51.PM.mov 
> Disconnect wallet actually not working 🙇🏼

Just to make sure I understand, the problem you are referring to is that the "Disconnect wallet" appears again after reloading the page, right?

If, so, at the moment, that is the configured behavior ( wagmi autoConnect is set to true in Providers.tsx). The user can still switch wallet by clicking "Disconnect wallet" and then Login (if they don't reload first).
If we change autoConnect to false, the wallet will remain disconnected after a reload. Then the user will need to press Login.

foolo:
> image
> 
> Messed up a bit 🙇🏼

@bigint Interesting, when are you seeing this? I can't seem to reproduce it (I tried resizing the browser, and switching to mobile view)

bigint:
> > image
> > Messed up a bit 🙇🏼
> 
> @bigint Interesting, when are you seeing this? I can't seem to reproduce it (I tried resizing the browser, and switching to mobile view)

This is in mobile view 🙇🏼 I think we need to change the design idea, is disconnect button is really valued? imo one button is good for ui

foolo:
> This is in mobile view 🙇🏼 I think we need to change the design idea, is disconnect button is really valued? imo one button is good for ui

@bigint Ok, I just did the same as Lenstube. Their disconnect button is much smaller though, just an icon.
These are the two possible options that I can think of:
1. (current solution) Give the user some way of disconnecting current wallet, (it has to be in logged-out mode), currently this is is what the disconnect button does.
2. Let the user select wallet before _every_ login.

foolo:
> This is in mobile view 🙇🏼 I think we need to change the design idea, is disconnect button is really valued? imo one button is good for ui

@bigint Ok, I just did the same as Lenstube, as decided (or at least as I though we decided 🙂 ).  I asked https://github.com/lensterxyz/lenster/pull/1613#issuecomment-1406741213 if I should use a disconnect button, same as Lenstube, and you said _"Yea we can do the same as lenstube"_, so I thought it was decided.

But of course we can always rethink the solution.

Then the question is:
If there is no disconnect button, and no dialog, then where/how will the user disconnect their wallet, to select another wallet? If there is no such option, we would need to let the user select wallet on _every_ login, which maybe would be bad UX.

bigint:
We need to rethink this and need to check how other platform does 🙇🏼

bigint:
Still i'm worried about the bundle size and vendor locking 🙇🏼, give me some time i will discuss with some frens reg this 🙇🏼

foolo:
> Still i'm worried about the bundle size and vendor locking 🙇🏼, give me some time i will discuss with some frens reg this 🙇🏼

What do you mean with vendor locking? You mean being locked to RainbowKit? I implemented this because you asked me specifically if I could take this issue :)

PR #1619, opened 2023-01-15
fix paragraph spacing

### LINKED ISSUE #1193: Line break not working as expected

### Current Behavior

its posting like this: image

### Expected Behavior

when i input this: 
image

### Steps To Reproduce

_No response_

Web

Edge

### Anything else?

_No response_

foolo:
## What does this PR do?

Fixes #1193
Problem: In editor, user could not see the difference between paragraph and newline. But in the published post, paragraph and newline are displayed differently (correctly).

So in this PR, the rendering of the published post is **not** changed. Instead, we fix the style in the editor, so that user sees how the post will actually look when published.

Note: Tailwind was setting all margins to 0, so therefor all elements, both `<p>` and `<br>`  had a margin of 0.

## Appearance before PR
Screenshot_2023-01-15_09-03-29 (copy 1)

## Appearance after PR
Screenshot_2023-01-15_09-44-33 (copy 1)

## Appearance of published post (not changed with this PR)
Screenshot_2023-01-15_09-54-44

## Type of change

## How should this be tested?

Open editor,
Type a line
Create new paragraph by pressing Enter
Observe large spacing
Type two more lines, separated by Shift+Enter
Observe smaller spacing

### DIFF
--- a/apps/web/src/components/Composer/Editor/index.tsx
+++ b/apps/web/src/components/Composer/Editor/index.tsx
@@ -40,7 +40,7 @@ const Editor: FC = () => {
       <EmojiPickerPlugin />
       <ToolbarPlugin />
       <RichTextPlugin
-        contentEditable={<ContentEditable className="px-5 block my-4 min-h-[65px] overflow-auto" />}
+        contentEditable={<ContentEditable className="px-5 block my-4 min-h-[65px] overflow-auto editor" />}
         placeholder={
           <div className="px-5 absolute top-[65px] text-gray-400 pointer-events-none whitespace-nowrap">
             <Trans>What's happening?</Trans>
--- a/apps/web/src/styles.css
+++ b/apps/web/src/styles.css
@@ -187,3 +187,7 @@ div[data-lexical-editor='true'] {
 #typeahead-menu {
   @apply z-20;
 }
+
+.editor p:not(:last-child) {
+	margin-bottom: 1em;
+}


bigint:
@foolo IMO this should be fixed in interweave side, the composer has expected behaviour

bigint:
Close T-3674

foolo:
> the composer has expected behaviour

Are you sure? :) Shouldn't the editor show whether the user typed a paragraph or a newline?
(Or maybe Interweve shouldn't show any difference either? In that case it would make sense, then we simply don't support paragraphs.)

bigint:
Yes, we have both line break (\n) and paragraph break (\n\n), this is bug in interweave, even with single \n interweave do \n\n

foolo:
> Yes, we have both line break (\n) and paragraph break (\n\n), this is bug in interweave, even with single \n interweave do \n\n

Actually, Interweave will show **one** line break '\n', if you do a newline in editor, (Shift+Enter), and **two** line breaks ('\n\n') if you do a paragraph in editor. I have investigated it quite thorougly, I would say :)

foolo:
Note from chat discussion: Editor needs to be modified too, so that it can only output one kind of newline (at the moment it can output either `\n` (Shift+Enter) or `\n\n` (Enter) , which means that in the output, there is no way of knowing how many newlines the editor actually showed).

PR #1622, opened 2023-01-15
feat: keep track of partially composed messages

### LINKED ISSUE #1007: Keep track of partially composed messages
### Summary

Unsent messages get cleared when you navigate to a different conversation, following https://github.com/lensterxyz/lenster/pull/1005.

We should probably keep track of the state of all unsent messages so that when the user navigates back to a conversation they can resume their writing.

foolo:
## What does this PR do?

Fixes #1007
Screen recording: https://user-images.githubusercontent.com/667227/212549701-e724b33e-e40b-4ffb-a59b-27b5d708d7e0.mp4

Note: Using the same localstorage key for all messages. It would also be possble to store each message as a separate key. Then no JSON parsing would be needed, and no need to read and write all messages to update just one message. But I think that there will probably not be too many messages on the same time, so I went with just one localstorage key.

## Type of change

## How should this be tested?

Type a message in one conversation, without sending the message. Navigate to another conversation and then back to the first one. Observe that the message is restored.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -10,7 +10,7 @@
     "lint": "eslint . --ext .js,.ts,.tsx"
   },
   "dependencies": {
-    "@apollo/client": "^3.7.3",
+    "@apollo/client": "^3.7.4",
     "@aws-sdk/client-s3": "^3.245.0",
     "@aws-sdk/client-sts": "^3.245.0",
     "@bundlr-network/client": "^0.9.11",
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -13,7 +13,7 @@
     "extract": "NODE_ENV=development lingui extract"
   },
   "dependencies": {
-    "@apollo/client": "^3.7.3",
+    "@apollo/client": "^3.7.4",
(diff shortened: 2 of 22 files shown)


bigint:
Do you think using zustand for this will be better?

bigint:
Close T-3672

bigint:
@foolo Do you think using zustand for this will be better?

foolo:
> @foolo Do you think using zustand for this will be better?

It might be, I'll try it!

bigint:
@foolo is it good to review now?

foolo:
@bigint I switched to zustand, and used the existing useMessagePersistStore store, which felt like a natural place.
But just tell if I should instead use a separate store.

foolo:
> @foolo is it good to review now?

@bigint Yes! Unless you see any problem with using useMessagePersistStore, see previous comment https://github.com/lensterxyz/lenster/pull/1622#issuecomment-1383933635

bigint:
**LGTM** :shipit: 

Rad LGTM

PR #1629, opened 2023-01-16
chore: move emoji picker call to tanstack query

bigint:


bigint:
Close [T-3360](https://avara.height.app/T-3360)

PR #1631, opened 2023-01-16
chore: remove success button variant across the platform

bigint:


bigint:
Close [T-3692](https://avara.height.app/T-3692)

PR #1635, opened 2023-01-17
fix: create only line breaks in editor

foolo:
## What does this PR do?

New PR for #1193

As concluded in chat with @bigint :

> fix Lexical so that it only support newlines (similar to twitter)

After this PR, the Lexical editor will **only** create line breaks (\n), regardless of whether user presses _Shift+Enter_ or _Enter_.
(It will not create paragraphs (exported as \n\n), which it did before on _Enter_)

## Type of change

## How should this be tested?

Write a post, try separating some lines with _Enter_ and some with _Shift+Enter_. Verify that in the resulting published post, the line breaks match how it looked in the editor.


bigint:
Link [T-3674](https://avara.height.app/T-3674)

bigint:
cc @harish-sethuraman for your eyes 🙇🏼

bigint:
**LGTM** :shipit: 

Rad LGTM

PR #1656, opened 2023-01-22
chore: remove cypress

bigint:


bigint:
Close [T-4225](https://avara.height.app/T-4225)

PR #1658, opened 2023-01-22
fix: null comparisons without type-checking operators may not work as intended

bigint:


bigint:
Close [T-4227](https://avara.height.app/T-4227)

PR #1667, opened 2023-01-23
fix: add follow modal in user profile

### LINKED ISSUE #1417: Add follow modal in user profile

foolo:
## What does this PR do?

Add a follow user modal when visiting profile with `?showFollowDialog=true` in url
Screenshot_2023-01-23_09-37-38

Fixes #1417
[Screen recording](https://user-images.githubusercontent.com/667227/213995892-51155e35-8f64-4270-9621-721ab3a8365d.mp4)

Behavior:
- If already following: don't show the modal (an alternative would be to show "you already follow this user")
- If not logged in, show the modal. If user clicks Follow, the login flow will start

## Type of change

## How should this be tested?

Open a profile, with query `?showFollowDialog=true` added to the url, for example:
https://lenster-b3i5tangl-foolo.vercel.app/u/olof3?showFollowDialog=true


apps/web/src/components/Profile/index.tsx
```diff
@@ -37,6 +44,38 @@ const ViewProfile: NextPage = () => {
...
+    const canFollow = followType !== 'RevertFollowModuleSettings' && !isFollowedByMe;
+    if (showFollowDialog && canFollow) {
+      setShowFollowModal(true);
+    }
+    setFollowing(isFollowedByMe);
+  }
+
+  // profile changes when user selects a new profile from search box
+  useEffect(() => {
+    if (profile != undefined) {
```
bigint, line 68:
```suggestion
    if (profile !== undefined) {
```

will `!profile` work here?

apps/web/src/components/Profile/index.tsx
```diff
@@ -49,10 +88,47 @@ const ViewProfile: NextPage = () => {
...
+                onClick={() => {
+                  setShowFollowModal(false);
+                }}
+                aria-label={t`Not now`}
+              >
+                <Trans>Not now</Trans>
+              </Button>
+            </span>
+          </div>
+        </div>
```
bigint, line 129: can we make this as a separate component?

bigint (changes requested):

apps/web/src/components/Profile/index.tsx
foolo: hmm, not sure why I wrote `if (profile != undefined)` and not simply `if (profile)`
i have updated the code

apps/web/src/components/Profile/index.tsx
foolo: Sure, now I moved the stuff that was inside <Modal> to a new component: apps/web/src/components/Profile/FollowDialog.tsx
You can have a look.

bigint:
Close T-2607

bigint:
Close T-2608

PR #1670, opened 2023-01-23
feat: protect dispatcher with IDKit

m1guelpf:
## What does this PR do?

Prompt users to enable IDKit after they configure the dispatcher, and adds an option for it on the dispatcher page for those who might have enabled before this was available.

## Type of change

## How should this be tested?

- [ ] On the dispatcher settings tab, verify that when `isIDKitPhoneVerified` flag is false the IDKit modal pops up
- [ ] On the dispatcher settings tab, verify that when `isIDKitPhoneVerified` flag is set to true, the IDKit modal doesn't pop up


bigint:
Link T-2927

bigint:
@m1guelpf thanks for the PR 🙏🏼 

What Lens API endpoint you used to do `yarn codegen`?

m1guelpf:
@bigint I didn't touch the endpoint, just ran `yarn codegen` without changing anything at all

bigint:
Ahh weird, I ran the same I don't see the changes, lemme ask Josh about this

bigint:
Not sure how it got removed for you!! I still see those mutations in production and testnet playground

apps/web/src/components/Settings/Dispatcher/ToggleDispatcher.tsx
```diff
@@ -22,6 +24,7 @@ interface Props {
 }
 
 const ToggleDispatcher: FC<Props> = ({ buttonSize = 'md' }) => {
+  const { data: idKitData } = useIsIdKitVerifiedQuery({});
```
bigint, line 27:
```suggestion
  const { data: idKitData } = useIsIdKitVerifiedQuery();
```

do we need to pass empty object?

apps/web/src/components/Settings/Dispatcher/ToggleDispatcher.tsx
```diff
@@ -94,23 +122,37 @@ const ToggleDispatcher: FC<Props> = ({ buttonSize = 'md' }) => {
...
-      onClick={toggleDispatcher}
+    <IDKitWidget
+      autoClose
+      actionId={IDKIT_ACTION_ID}
+      onSuccess={toggleDispatcher}
+      handleVerify={handleIDKitProof}
+      signal={currentProfile?.ownedBy}
+      copy={{
+        title: 'Lens Dispatcher',
+        heading: t`Verify your phone number for free gassless transactions`
```
bigint, line 133: make sure to run `yarn i18n:extract` at last 🙇🏼

apps/web/src/components/Settings/Dispatcher/ToggleDispatcher.tsx
```diff
@@ -85,6 +88,31 @@ const ToggleDispatcher: FC<Props> = ({ buttonSize = 'md' }) => {
     } catch {}
   };
 
+  const handleIDKitProof = async (result: ISuccessResult) => {
+    const response = await fetch('https://world-id-lens-bridge.vercel.app/api/v1/submit', {
```
bigint, line 92: Are we going to use this endpoint or will it comes under any other subdomain? also I think we can move the URL to constants.

apps/web/src/components/Settings/Dispatcher/ToggleDispatcher.tsx
```diff
@@ -85,6 +88,31 @@ const ToggleDispatcher: FC<Props> = ({ buttonSize = 'md' }) => {
...
+      return;
+    }
+
+    if (response.status === 400 && (await response.json()).code === 'already_verified') {
+      throw new Error(
+        'You have already verified this phone number with Lens. You can only verify one wallet with one phone number.'
+      );
+    }
+
+    throw new Error('Something went wrong. Please try again.');
```
bigint, line 113: I think better can use `toast.error` instead of throwing an error at the app level because we don't want to log these kinds of errors in our datadog aka these are for users not for system.

bigint (changes requested):
@m1guelpf left some trivial comments 🙇🏼

apps/web/src/locales/es/messages.po
```diff
@@ -1620,6 +1620,10 @@ msgstr ""
 msgid "Profile updated successfully!"
 msgstr ""
 
+#: src/components/Settings/Dispatcher/ToggleDispatcher.tsx
+msgid "Verify your phone number for free gassless transactions"
+msgstr "Verifica tu número de teléfono para que cubramos el gas de tus transacciones"
```
bigint, line 1625: Thanks for translating as well 😅🙏🏼

apps/web/src/components/Settings/Dispatcher/ToggleDispatcher.tsx
bigint: Ahh cool then 👍🏼

apps/web/src/components/Settings/Dispatcher/ToggleDispatcher.tsx
```diff
@@ -22,14 +24,54 @@ interface Props {
...
+        })
+      });
+
+      if (response.ok) {
+        return;
+      }
+
+      if (response.status === 400 && (await response.json()).code === 'already_verified') {
+        throw new Error(
+          'You have already verified this phone number with Lens. You can only verify one wallet with one phone number.'
```
bigint, line 55: If we localize this will it work? as you mentioned this will be displayed inside the widget 🙇🏼

apps/web/src/components/Settings/Dispatcher/ToggleDispatcher.tsx
```diff
@@ -22,14 +24,54 @@ interface Props {
...
   const setUserSigNonce = useAppStore((state) => state.setUserSigNonce);
   const currentProfile = useAppStore((state) => state.currentProfile);
   const canUseRelay = currentProfile?.dispatcher?.canUseRelay;
 
+  const { setOpen: setIDKitOpen } = useIDKit({
+    autoClose: true,
+    enableTelemetry: true,
+    actionId: IDKIT_ACTION_ID,
+    handleVerify: async (result: ISuccessResult) => {
+      const response = await fetch(IDKIT_BRIDGE, {
```
bigint, line 38: we always prefer using axios with tanstack query 🙇🏼, it would be great if we migrate to that 🚀

bigint (changes requested):

apps/web/src/components/Settings/Dispatcher/ToggleDispatcher.tsx
m1guelpf: I'm not super familiar with how the translation system Lenster uses works, but if it'll just return a string in the proper language it should.

apps/web/src/components/Settings/Dispatcher/ToggleDispatcher.tsx
m1guelpf: Sure! I'm not super familiar with tanstack query, would love if you could provide an example on its usage with a POST endpoint somewhere in lenster I can look at

apps/web/src/components/Settings/Dispatcher/ToggleDispatcher.tsx
bigint: Ahh ok, then leave as it is, I'll experiment it later 🙇🏼

apps/web/src/components/Settings/Dispatcher/ToggleDispatcher.tsx
bigint: We don't have any POST request in lenster so far in tanstack, but here is the simple example how to do

https://medium.com/analytics-vidhya/how-to-post-and-fetch-data-using-react-query-4c3280c0ef96

m1guelpf: Looking at the linked article, my understanding is that there's no way to get the response inline in the function, and you handle success/error states by adding `onSuccess`/`onError` callbacks on the widget.

This does unfortunately not work in this scenario, since `handleVerify` is ran by the IDKit widget, and we need to keep the state (or throw the error) inside of the function so IDKit can display the error to the user or know the validation has succeeded.

I can move `fetch` to `axios` however, if that fits the code conventions better, lmk

apps/web/src/components/Settings/Dispatcher/ToggleDispatcher.tsx
```diff
@@ -85,32 +94,80 @@ const ToggleDispatcher: FC<Props> = ({ buttonSize = 'md' }) => {
...
+        signal={currentProfile?.ownedBy}
+        copy={{
+          title: 'Lens Dispatcher',
+          heading: t`Verify your phone number to increase gassless limits`
+        }}
+      >
+        {({ open }) => {
+          (openIDKit as MutableRefObject<() => void>).current = open;
+          return <span />;
+        }}
```
m1guelpf, line 145: new release of IDKit (tomorrow) makes this much cleaner

apps/web/src/components/Settings/Dispatcher/index.tsx
```diff
@@ -45,6 +79,38 @@ const DispatcherSettings: FC = () => {
...
+            <IDKitWidget
+              enableTelemetry
+              methods={['phone', 'orb']}
+              actionId={IDKIT_ACTION_ID}
+              handleVerify={handleIDKitVerify}
+              signal={currentProfile?.ownedBy}
+              copy={{
+                title: 'Lens Dispatcher',
+                heading: t`Verify your phone number to increase gassless limits`
+              }}
```
m1guelpf, line 99: @bigint Since these props are duplicated across two files, what would be the most appropriate place to store a "wrapper component"?

bigint:
Closing this for now! we can re-open the same PR when we finish the spec internally 🙇🏼

PR #1673, opened 2023-01-24
feat: add pwa offline page

bigint:


bigint:
Close [T-3739](https://avara.height.app/T-3739)

PR #1689, opened 2023-01-27
fix: minor typos

omahs:
Fix: minor typos


bigint:
These doesn't change any meaning either not a grammar error as well 🙇🏼

omahs:
@bigint These are grammar errors, yeah small ones.
– The subject of the sentence is **The Site** which is singular, so the verb should be in the singular form.
– The indefinite article **a** is redundant when used with the uncountable noun **content** in the second sentence.
– Verb is missing in last sentence.

bigint:
This should not be done in .po files

Check this https://github.com/lensterxyz/lenster/blob/main/docs/internationalization.md

Closing this now 🙇🏼

PR #1726, opened 2023-02-03
chore: remove babel

bigint:


foolo:
Seems to work! At least the parts I could think of needed testing.
I tested that lingui extract and compile work, and that the translation on the site works, and I can't see anything wrong with the code.

bigint:
Thanks for testing @foolo 🙏🏼 

Vercel is down badly for 3+ hrs, will merge once this is deployed in previews 🚀

PR #1754, opened 2023-02-08
fix: remove paragraphs from editor

foolo:
## What does this PR do?

Fixes new problem as described in comment https://github.com/lensterxyz/lenster/issues/1193#issuecomment-1420805409

This works by modifying the editor's content tree, replacing all paragraphs with newlines, before the content is exported to markdown

See also https://lexical.dev/docs/concepts/editor-state#updating-state

## Type of change

## How should this be tested?

- Open publication editor, **paste** the content below and publish the post

```
aaaaaa
bbbb

foo
bar
```

- Verify that the published post looks the same

- Repeat the above steps, but typing the content by hand, instead of pasting it


bigint:
Looks too complex :(

bigint:
Looks too complex :(

foolo:
> Looks too complex :(
@bigint Yeah, I agree it is a bit complex. But on the other hand it's the only solution so far that actually works 🙂
It's not for me to decide whether it's too complex though.
We could also try to think of a new, different solution.

bigint:
cc @harish-sethuraman 🙇🏼

harish-sethuraman:
Isn't this solved yet (I remember replacing two line feed with one)? I don't think we have to deal with this given that lexical is having `\n\n` as block separator? 

their code here:
https://github.com/facebook/lexical/blob/3228323a2e32245386dacc0327d844de6390e5a6/packages/lexical-markdown/src/MarkdownExport.ts#L55

foolo:
> Isn't this solved yet (I remember replacing two line feed with one)? I don't think we have to deal with this given that lexical is having `\n\n` as block separator?
> 
> their code here: https://github.com/facebook/lexical/blob/3228323a2e32245386dacc0327d844de6390e5a6/packages/lexical-markdown/src/MarkdownExport.ts#L55

The above fix (replacing \n\n with \n) unfortunately introduced a new error: when making an _empty line by hand_, that empty line disappeared

@bigint removed the that fix again after an update of Lexical, which was allegedly working. But now there is a new problem:
You can reproduce it via comment: https://github.com/lensterxyz/lenster/issues/1193#issuecomment-1420805409
Try it :)
This PR fixes the problem.

harish-sethuraman:
I see that when I used the `notes` app in Mac to copy paste the following two new lines appeared instead of one in your PR. But when I copied from github it works fine. We need to fix that as well? we dont want two new lines instead of one while pasting from notes? We might want

```
Paste
This
And

See
magic
```

From what I found is when we copy paste from notes app we get an line break node
Screenshot 2023-02-14 at 1 03 19 PM

If we use some other app we dont get that
Screenshot 2023-02-14 at 1 02 14 PM

foolo:
@harish-sethuraman Thanks for testing! Agree, the new bug you found is not good.
We can probably tweak this PR to cover that case. But, of course, this is not really a long term solution, since it's quite hackish, and there's a big chance that even more bugs will show up for all kinds of tricky cases.

I discussed with @bigint before about that the long-term solution is to use Lexical also for rendering (in read-only mode), which will be even more necessary if we want to support editing of existing posts.

PR #1772, opened 2023-02-09
feat: image cropper

### LINKED ISSUE #76: Add image cropper
ENG-194

foolo:
## What does this PR do?

Fixes #76
Close ENG-194

Add cropping feature when uploading a profile picture

## Type of change

## How should this be tested?

Go to Profile settings > Upload avatar, select a file, crop it and press Upload and save, verify the new profile picture


bigint:
I'm looking something like twitter or github has 🙇🏼 

the problem with current implementation is, if the image size is very big the crop area become too small, aka very difficult to crop

image

foolo:
@bigint Probably I can improve the UX a lot by writing custom pan-and-zoom logic. Then we don't need the third-party library.

bigint:
Yup that would be super cool 🚀

bigint:
@foolo are we taking this forward? IMO we can focus 1 PR at a time, first we can finish this and we can move to rainbow 🙇🏼  WDYT?

foolo:
> @foolo are we taking this forward? IMO we can focus 1 PR at a time, first we can finish this and we can move to rainbow 🙇🏼 WDYT?

Yep, working on this one at the moment :) Agree, better to focus on one PR at a time. The Rainbow wallet PR is probably closer to done than this one, but for me it doesn't matter which one to prioritize.

Regarding this one, I'm working on improving the UX. As you mentioned, the previous version (using react-easy-crop package) had bad UX for very tall or wide images.

Attached a video below, of where I'm at right now.
A question regarding desired features: Twitter shows the cropper it in a popup-modal, and it also has a zoom-slider below the image. Do we want any of those features?

https://user-images.githubusercontent.com/667227/218449486-4f7bf98d-ab6d-49cf-9762-ecea16128582.mp4

foolo:
@bigint I updated the behavior a bit. You can take a look on the deployment and share feedback.

bigint:
Functionality really looks great 👏🏼

bigint:
Can we make this for cover image as well? (1500x500)

bigint:
Also IMO we can move the cropper into a modal 🙇🏼

foolo:
> Functionality really looks great 👏🏼

Great, thanks! :+1: 

> Can we make this for cover image as well? (1500x500)

@bigint, Yep, I think it should be easy. But if it turns out to be more complicated than expected, maybe we can do it in a following pr?

> Also IMO we can move the cropper into a modal 🙇🏼

Yes, that should be no problem. I'll look at it.

foolo:
@bigint I moved the cropper to a modal.
Right now, though, I'm blocked by two things:
- Upload to IPFS not working in devmode, (happens also on main branch). Seems to be some credentials problem. I haven't been able to figure it out.
Screenshot_2023-02-12_10-45-45
https://github.com/lensterxyz/lenster/blob/image-cropper-76/apps/web/src/lib/uploadToIPFS.ts#L53

- Can't connect to testnet via mobile, which is needed to test the cropper behavior on mobile.

bigint:
I'll address this issue asap @foolo, I'll fix tomorrow 🙇🏼

foolo:
> I'll address this issue asap @foolo, I'll fix tomorrow 🙇🏼

Nice, I hope you can find a solution. I would really need it to complete this issue :) It seems like the reason is that EVER_ACCESS_KEY and EVER_ACCESS_SECRET are not set, so `handleRequest()` in `packages/workers/sts-generator/src/index.ts` returns `{ success: false, message: 'Something went wrong!' }`

foolo:
> I'll address this issue asap @foolo, I'll fix tomorrow 🙇🏼

@bigint I managed to work around the problem with uploadToIPFS, so the fix isn't needed for this PR.
I opened the PR for review. You can take a look.
Not sure whether it's best to include also the cover picture in this PR. It's not there now at least.

foolo:
> We can move cover picture to next PR 🙇🏼, Id always suggest to have a drag component so it will be easy in mobile too
> 
> Screenshot 2023-03-07 at 12 32 17 PM

@bigint Sounds good, thanks for feeback 👍 I'll add a zoom slider.

foolo:
@bigint Btw, if you have some opinion regarding the slider (or anything else in this issue), please share. Just to avoid too many iterations back and forth :) For example if we should prefer some specific library for the slider component.

foolo:
@bigint I added a slider. Check it out!

foolo:
Another note: This implementation is based on [react-easy-crop](https://github.com/ValentinH/react-easy-crop), which is MIT licence, so maybe we need to put that [LICENSE](https://github.com/ValentinH/react-easy-crop/blob/main/LICENSE) somewhere in Lenster repo?

foolo:
> @foolo can't we use their library directly instead of bringing their code to our codebase?

@bigint I have changed code a lot to improve the UX/math/logic for zooming, aspect ratio, resolution, etc, so it would be impossible to use their library without losing all that.

But  I'm a bit confused. I used the library directly in the first version of this PR, and you wrote:

> the problem with current implementation is, if the image size is very big the crop area become too small, aka very difficult to crop

So I suggested that I fork the code and improve the logic UX, to which you replied

> Yup that would be super cool

And that's what I have now done :)

bigint:
haha lol makes sense 😅 

for license we can add it in the image-cropper folder itself, I suggest to move that into packages 🙇🏼 instead of app level

foolo:
> for license we can add it in the image-cropper folder itself, I suggest to move that into packages 🙇🏼 instead of app level

Ok, I'll do that!

foolo:
> for license we can add it in the image-cropper folder itself, I suggest to move that into packages 🙇🏼 instead of app level

@bigint, Ok, I'll do that!

bigint:
Close T-3

packages/image-cropper/LICENSE.txt
```diff
@@ -0,0 +1,21 @@
+MIT License
```
bigint, line 1: we can remove `.txt` extension

packages/image-cropper/cropUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+  ctx.drawImage(image, 0, 0);
+
+  const data = ctx.getImageData(pixelCrop.x, pixelCrop.y, pixelCrop.width, pixelCrop.height);
+
+  canvas.width = pixelCrop.width;
+  canvas.height = pixelCrop.height;
+  ctx.putImageData(data, 0, 0);
+
+  return canvas;
+}
```
bigint, line 88: Can we make all arrow function?

packages/image-cropper/README.md
```diff
@@ -0,0 +1,3 @@
+# Image cropper
```
bigint, line 1: Amazing work on migrating to packages 👏🏼

apps/web/src/lib/uploadToIPFS.ts
```diff
@@ -49,7 +49,8 @@ const uploadToIPFS = async (data: any): Promise<LensterAttachment[]> => {
     );
 
     return attachments;
-  } catch {
+  } catch (error) {
+    console.error(error);
```
bigint, line 53: we don't want to throw errors on prod!! maybe in future we can make a logger lib that yells only on development

packages/image-cropper/ImageCropper.tsx
```diff
@@ -0,0 +1,458 @@
...
+import {
+  computeCroppedArea,
+  getDistanceBetweenPoints,
+  getMidpoint,
+  restrictPosition,
+  restrictValue
+} from './cropUtils';
+import type { Area, MediaSize, Point, Size } from './types';
+
+export type CropperProps = {
```
bigint, line 14: better we can use interface everywhere instead of types 🙇🏼

packages/image-cropper/ImageCropper.tsx
```diff
@@ -0,0 +1,458 @@
...
+  hasWheelJustStarted: boolean;
+};
+
+type GestureEvent = UIEvent & {
+  scale: number;
+  clientX: number;
+  clientY: number;
+};
+
+class ImageCropper extends React.Component<CropperProps, State> {
```
bigint, line 36: we can just import component instead of importing full react package (tree shaking)

bigint: In all places 🙇🏼

packages/image-cropper/ImageCropper.tsx
```diff
@@ -0,0 +1,458 @@
...
+  rafPinchTimeout: number | null = null;
+  wheelTimer: number | null = null;
+  currentDoc: Document = document;
+  currentWindow: Window = window;
+
+  state: State = {
+    hasWheelJustStarted: false
+  };
+
+  componentDidMount() {
```
bigint, line 62: is it possible to migrate to functional component?

apps/web/src/components/Shared/ChooseFile.tsx
```diff
@@ -11,6 +11,9 @@ const ChooseFile: FC<ChooseFileProps> = ({ onChange }) => {
       type="file"
       accept=".png, .jpg, .jpeg, .gif"
       onChange={onChange}
+      onClick={(event) => {
+        (event.target as HTMLInputElement).value = '';
```
bigint, line 15: we can do something like this right? no?

`onClick={(event: HTMLInputElement) => {`

apps/web/src/components/Settings/Profile/Picture.tsx
```diff
@@ -106,82 +107,131 @@ const Picture: FC<PictureProps> = ({ profile }) => {
...
+
+  function readFile(file: Blob): Promise<string> {
+    return new Promise((resolve) => {
+      const reader = new FileReader();
+      reader.addEventListener('load', () => resolve(reader.result as string), false);
+      reader.readAsDataURL(file);
+    });
+  }
+
+  const onFileChange = async (evt: ChangeEvent<HTMLInputElement>) => {
```
bigint, line 172: evt => event

apps/web/src/components/Settings/Profile/Picture.tsx
```diff
@@ -106,82 +107,131 @@ const Picture: FC<PictureProps> = ({ profile }) => {
...
+        await createSetProfileImageURITypedData({
+          variables: {
+            options: { overrideSigNonce: userSigNonce },
+            request
+          }
+        });
       }
+      setAvatarDataUrl(dataUrl);
+    } catch (error) {
+      console.error(error);
```
bigint, line 153: here too 🙇🏼

apps/web/src/components/Settings/Profile/ImageCropperController.tsx
```diff
@@ -0,0 +1,63 @@
+import 'rc-slider/assets/index.css';
+
+import { ZoomInIcon, ZoomOutIcon } from '@heroicons/react/outline';
+import Cropper from 'image-cropper/ImageCropper';
+import type { Area, Point, Size } from 'image-cropper/types';
+import Slider from 'rc-slider';
+import type { Dispatch, FC } from 'react';
+import { useRef, useState } from 'react';
+
+interface Props {
```
bigint, line 10: Props => ImageCropperControllerProps

bigint (changes requested):

bigint:
@foolo left some comments 🙇🏼

packages/image-cropper/README.md
foolo: Thanks :) It became a bit easier when I found that there were actually two existing packages with code in them (`utils` and `data`) so I had somewhere to start from. I hope that everything in package.json and tsconfig.json makes sense.

apps/web/src/lib/uploadToIPFS.ts
foolo: I assume you mean we should remove the log here. I'll remove it! I think your suggestion about a logger library is important to follow up on. It can save many hours of troubleshooting if there actually is a bug.

My reasoning behind keeping the log output was that I think that in general, a catch clause should always be handled, even if it's only with log, be it on client or server side.

apps/web/src/components/Shared/ChooseFile.tsx
foolo: I tried your suggestion but the function signature for `onClick` is `(event: MouseEvent<HTMLInputElement, MouseEvent>)`.  I couldn't get it working, but further suggestions appreciated, if you can find another way.

Regarding typecast, it seems that it is needed here. I think this is relevant: https://stackoverflow.com/a/57701995/961254

I'll add a comment though, because it's not obvious what this line does :)

packages/image-cropper/ImageCropper.tsx
foolo: Sure, I'll update. For some types, there seems to be a distinction  between with and without `React.`, they point to different libs, so i have kept it as-is in those places.
For example `MouseEvent` vs `React.MouseEvent`:
`(e: MouseEvent | React.MouseEvent | GestureEvent)`

packages/image-cropper/cropUtils.ts
foolo: done!

packages/image-cropper/ImageCropper.tsx
foolo: I tried it before, and I think there were some problems, or seemed like a lot of work. But I don't remember exactly, so I'll look at it again.

packages/image-cropper/README.md
bigint: 🙌🏼

packages/image-cropper/ImageCropper.tsx
bigint: It's fine then, its not a big type issue 🙇🏼

bigint: We can import them carefully 🙇🏼

apps/web/src/components/Shared/ChooseFile.tsx
bigint: It's fine then, its not a big type issue 🙇🏼

packages/image-cropper/ImageCropper.tsx
bigint: Cool not urgent

bigint:
One last thing 🙇🏼 

image

Can we make the contents inside the modal full width and also align save button to right 🙇🏼 🙏🏼

foolo:
> Can we make the contents inside the modal full width and also align save button to right 🙇🏼 🙏🏼

@bigint At the moment, although the size of the cropper is configurable, it does not support _dynamic_ resizing.
So there are some different options:

1. Add support for dynamic resizing. Will take some time to implement.
 
2. Just shrink the modal:
Screenshot_2023-03-13_12-56-11

3. Increase size of cropper. But then there would need to be some logic to only increase the size on desktop, and not on mobile.
Screenshot_2023-03-13_12-55-23

4. Center the cropper, which may look a bit better:
Screenshot_2023-03-13_13-01-52

bigint:
Amazing work 🚀  bls resolve conflicts 🙏🏼

bigint (approved):
**LGTM** :shipit: 

Rad LGTM

foolo:
> Amazing work rocket bls resolve conflicts 🙏🏼

Thanks :) I've resolved the conflicts.

PR #1829, opened 2023-02-20
adding meta description and og:description

abhiram11:
## What does this PR do?
This PR adds `<meta>` tags of `description` (for search engines like Google) and `og:description` (for open-graph/social media networks). The content for both the tags is the same, and copied exactly from Lenster's profile from Github/Twitter, etc.

Fixes # None, but the issue image is displayed as follows:

lenster

## Type of change

## How should this be tested?

- N/A 😢

## Checklist
All seems clear.

**P.S. I am new to open-source contribution, so I am all ears to feedback!**

### DIFF
--- a/apps/web/src/pages/_document.tsx
+++ b/apps/web/src/pages/_document.tsx
@@ -22,8 +22,14 @@ class LensterDocument extends Document {
           <meta name="apple-mobile-web-app-title" content="Lenster" />
 
           {/* SEO */}
-          <meta name="description" content="Lenster is a decentralized, and permissionless social media app built with Lens Protocol." />
-          <meta property="og:description" content="Lenster is a decentralized, and permissionless social media app built with Lens Protocol." />
+          <meta
+            name="description"
+            content="Lenster is a decentralized, and permissionless social media app built with Lens Protocol."
+          />
+          <meta
+            property="og:description"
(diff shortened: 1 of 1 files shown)


bigint:
Thanks for the PR 🙏🏼 

Here the description gonna get conflicted with our prerender service for posts and profile

image

abhiram11:
I see! Apologies for not testing it thoroughly. I can see two paths moving forward with this:

- We add a default and static `meta description` for the base URL `lenster.xyz` (and all non-custom URLs like _Contact_ or upcoming _About_ Pages) in the prerender as well, or add that as done in this PR and let the code-base overwrite the meta tags based on the profile/post URL (I am not fully-versed with SSG or other features, if I am not seeing some obvious problem with this approach, please let me know 😅).
- Should I close this PR if there is no solution? I would love to show some metadata on search engines for Lenster though! 😄

bigint:
@abhiram11 It's not good idea to add all URLs under prerender Cloudflare worker, which will increase the bill by 100x, I will think of different idea 🙇🏼 

Closing this now 🙏🏼 appreciate your work 🕺🏼

abhiram11:
I did not know that! Thanks for enlightening me as well! 💸🙏🏻

PR #1832, opened 2023-02-20
chore: migrate to ar:// instead of raw gateway url

bigint:
Close T-4068


sasicodes:
We should revert this I think, the profile metadata is not handled in BE.

PR #1921, opened 2023-03-09
feat: add tooltips

foolo:
## What does this PR do?

Improve user experience of first-time visitors by adding tooltips for buttons that have no text on them.
(For example, from my own experience I remember being a bit confused by the notification icon when i visited the site the first time).

Also changed one tooltip text to "New conversation" which is what it does, (and it's also the title of the dialog that pops up when clicking the button).

Screenshot_2023-03-09_10-49-02
Screenshot_2023-03-09_10-49-10
Screenshot_2023-03-09_10-50-48
Screenshot_2023-03-09_10-49-58
Screenshot_2023-03-09_10-51-14

## Type of change

### DIFF
--- a/apps/web/src/components/Messages/MessageIcon.tsx
+++ b/apps/web/src/components/Messages/MessageIcon.tsx
@@ -1,6 +1,8 @@
+import { Tooltip } from '@components/UI/Tooltip';
 import useXmtpClient from '@components/utils/hooks/useXmtpClient';
 import { MailIcon } from '@heroicons/react/outline';
 import conversationMatchesProfile from '@lib/conversationMatchesProfile';
+import { t } from '@lingui/macro';
 import type { DecodedMessage } from '@xmtp/xmtp-js';
 import { fromNanoString, SortDirection } from '@xmtp/xmtp-js';
 import Link from 'next/link';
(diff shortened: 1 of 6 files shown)


foolo:
> Hmm thanks for the PR! but IMO its not required 🙇🏼 its annoying when seeing tooltip again and again 🙇🏼

@bigint Ahh, OK, no problem. Maybe also a matter of taste. Personally I like tooltips for everything :) 
Regarding annoyance, I guess that the "withDelay" solves part of that, so for someone who already knows the button, they won't see the popup, because they click it immediately.
Actually I was most confused with the Notifications icon. The others I added just as a nice-to-have.

bigint:
We can think about a new icon for notification 🙇🏼

foolo:
> We can think about a new icon for notification 🙇🏼

@bigint, That's a good idea. I checked what  icon is used on some other sites (Twitter, Facebook, Youtube, Reddit,)  and they all use a bell icon. So it seem like that should be the most intuitive for a new user. For example, Reddit and Youtube:
Screenshot_2023-03-10_09-21-44

Screenshot_2023-03-10_09-11-14

I also investigated a bit how they use tooltips on other sites:
On Twitter, Facebook and Reddit, they have tooltips for all icon-only buttons.
But on Instagram, there are no tooltips at all.
These are just observations. Of course, Lenster doesn't have to do the same as any other site 🙂

bigint:
Yea we can consider this later on lenster redesign 🙇🏼 

But we can change the icon now 🔔 

Closing this PR 🙏🏼

PR #2113, opened 2023-03-20
Image cropper cover 2110

### LINKED ISSUE #2110: Image cropper for cover picture
### Summary

Use the image cropper when uploading a cover picture on profile settings page.
Follow-up from #76.

foolo:
## What does this PR do?

Use the image cropper when uploading a cover picture on profile settings page.
Add dynamic resizing to image cropper, to fit the with of the modal, which may change when user resizes browser, etc.

Fixes #2110

## Type of change

## How should this be tested?

Visit Profile settings > Profile > Cover > Choose file
In the modal that pops up, crop the image and click Upload. Modal will close.
Click Save
Go to profile page and verify that the cover image is updated with the cropped image.

### DIFF
--- a/apps/web/src/components/Settings/Profile/Picture.tsx
+++ b/apps/web/src/components/Settings/Profile/Picture.tsx
@@ -195,7 +195,7 @@ const Picture: FC<PictureProps> = ({ profile }) => {
         <div className="space-y-3">
           <div>
             <Image
-              className="rounded-lg"
+              className="max-w-xs rounded-lg"
               src={avatarDataUrl || profilePictureIpfsUrl}
               alt={t`Profile picture crop preview`}
             />
--- a/apps/web/src/components/Settings/Profile/Picture.tsx
+++ b/apps/web/src/components/Settings/Profile/Picture.tsx
@@ -7,13 +7,13 @@ import { Spinner } from '@components/UI/Spinner';
 import { PencilIcon } from '@heroicons/react/outline';
 import { Mixpanel } from '@lib/mixpanel';
 import onError from '@lib/onError';
+import uploadCroppedImage, { readFile } from '@lib/profilePictureUtils';
 import splitSignature from '@lib/splitSignature';
-import uploadToIPFS from '@lib/uploadToIPFS';
 import { t, Trans } from '@lingui/macro';
 import { LensHub } from 'abis';
 import { AVATAR, ERROR_MESSAGE, LENSHUB_PROXY, SIGN_WALLET } from 'data/constants';
 import { getCroppedImg } from 'image-cropper/cropUtils';
-import type { Area, Size } from 'image-cropper/types';
+import type { Area } from 'image-cropper/types';
 import type { MediaSet, NftImage, Profile, UpdateProfileImageRequest } from 'lens';
 import {
   useBroadcastMutation,
@@ -107,19 +107,6 @@ const Picture: FC<PictureProps> = ({ profile }) => {
     }
   };
 
-  const uploadImage = async (image: HTMLCanvasElement): Promise<{ ipfsUrl: string; dataUrl: string }> => {
-    const blob = await new Promise((resolve) => image.toBlob(resolve));
-    let file = new File([blob as Blob], 'cropped_image.png', { type: (blob as Blob).type });
-    const attachment = await uploadToIPFS([file]);
-    if (attachment[0]?.item) {
-      const ipfsUrl = attachment[0].item;
-      const dataUrl = image.toDataURL('image/png');
-      return { ipfsUrl, dataUrl };
-    } else {
-      throw new Error('uploadToIPFS failed');
-    }
-  };
-
   const uploadAndSave = async () => {
     if (!currentProfile) {
       return toast.error(SIGN_WALLET);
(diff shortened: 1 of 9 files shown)


foolo:
@bigint Opening as draft because I'm not sure about what workflow is the best choice. It's a bit different than Profile picture (where it uploads and saves at the same time). On cover picture, first you upload, and then click Save, to save the form together with the other values (profile name, bio, etc.).

https://user-images.githubusercontent.com/667227/226363132-9672d49f-7895-4da4-bb32-7ee7dce90af1.mp4

bigint:
@foolo lemme know when its ready for review 🙇🏼

foolo:
> @foolo lemme know when its ready for review 🙇🏼

Sure! I opened it as a draft because I was not sure about what user workflow is the best choice, as mentioned here https://github.com/lensterxyz/lenster/pull/2113#issuecomment-1476288433.
So just tell in case you have any opinion on that.

bigint (changes requested):

foolo:
> @foolo lemme know when its ready for review 🙇🏼

@bigint, It's ready for review. And as a bonus it will solve the display error: https://github.com/lensterxyz/lenster/issues/2539

apps/web/src/components/Settings/Profile/Picture.tsx
```diff
@@ -128,7 +115,8 @@ const Picture: FC<PictureProps> = ({ profile }) => {
 
     try {
       setUploading(true);
-      let { ipfsUrl, dataUrl } = await uploadImage(croppedImage);
+      const ipfsUrl = await uploadCroppedImage(croppedImage);
+      const dataUrl = croppedImage.toDataURL('image/png');
```
saagu335, line 119: Are we converting any type of image to png?

apps/web/src/components/Settings/Profile/Profile.tsx
```diff
@@ -172,7 +172,7 @@ const ProfileSettingsForm: FC<ProfileSettingsFormProps> = ({ profile }) => {
       const id = await uploadToArweave({
         name,
         bio,
-        cover_picture: cover ? cover : null,
+        cover_picture: coverIpfsUrl ? coverIpfsUrl : null,
```
saagu335, line 175: will `coverIpfsUrl ?? null` works?

apps/web/src/components/Settings/Profile/Profile.tsx
```diff
@@ -215,76 +215,151 @@ const ProfileSettingsForm: FC<ProfileSettingsFormProps> = ({ profile }) => {
...
   return (
-    <Card className="p-5">
-      <Form
-        form={form}
-        className="space-y-4"
-        onSubmit={({ name, location, website, twitter, bio }) => {
-          editProfile(name, location, website, twitter, bio);
-        }}
+    <>
+      <Modal
```
saagu335, line 263: Add modal to last 🙇  instead of start in the component

apps/web/src/components/Settings/Profile/Profile.tsx
```diff
@@ -215,76 +215,151 @@ const ProfileSettingsForm: FC<ProfileSettingsFormProps> = ({ profile }) => {
...
+              <img className="h-5 w-5" src="/pride.svg" alt="Pride Logo" />
+              <span>
+                <Trans>Celebrate pride every day</Trans>
+              </span>
+            </div>
+            <div className="flex items-center space-x-2">
+              <Toggle on={pride} setOn={setPride} />
+              <div className="lt-text-gray-500">
+                <Trans>Turn this on to show your pride and turn the {APP_NAME} logo rainbow every day.</Trans>
+              </div>
```
saagu335, line 349: Is the indentation correct? It looks weird on GitHub

Screenshot 2023-04-16 at 10 51 15 AM

saagu335 (changes requested):
Make sure to run `pnpm i18n:extract` 🙏

apps/web/src/components/Settings/Profile/Picture.tsx
foolo: Yes, but actually it seems that the backend converts everything to WebP format when it's uploaded, (if you download an avatar image or a cover image, the format is WebP). So I guess that PNG is to be seen as an intermediate format here, for uploading. I have not experimented with different formats though.

bigint: Makes sense 👍🏼

apps/web/src/components/Settings/Profile/Profile.tsx
foolo: Looks good in source code, and `pnpm lint` passes, so I think it's OK . Probably just looks like that on Github.
Btw, I usually use the "Hide whitespace" feature in Github diff settings. Then it won't show the above part at all, because it has only changed indentation.

apps/web/src/components/Settings/Profile/Profile.tsx
foolo: Not as a one-line change, because, `??` checks for nullish values, but `?` checks for "falsy" values. So they will work differently for the empty string `''`, which is the default value for `coverIpfsUrl`.
We could of course change to:
`const [coverIpfsUrl, setCoverIpfsUrl] = useState<string | null>(null);` (and from `coverIpfsUrl` to `coverIpfsUrl ?? ''` wherever it's used) but I don't know if it's better?

apps/web/src/components/Settings/Profile/Profile.tsx
foolo: done!

foolo:
> Make sure to run `pnpm i18n:extract`

done, thanks for the reminder!

PR #2149, opened 2023-03-22
feat: add image attachment support to chat

nakajima:
This PR adds support for sending E2E encrypted images in messages.

It uses the newly introduced `Attachment` and `RemoteAttachment` content types which you can read more about here: https://xmtp.org/blog/attachments-and-remote-attachments.

https://user-images.githubusercontent.com/483/227057859-b5dd6059-cdd5-4224-81ac-22d97029afe9.mov

I had to change some of the hardcoded height values that were in the messages area before to accommodate the composer growing when you're about to send an image. I also added a slight background color to the composer because it looked confusing to have a bunch of white space when you had an attachment preview in there. Feel free to change any of that if it doesn't work for you.

## Type of change

## How should this be tested?

You should be able to add an image by clicking the photo button in the chat view.


bigint:
Can we make like a regular chat attachment to upload and load seamlessly? 🙇🏼 

We don't want to show attachment name, load button etc

Also is it possible to send text along with attachments?

nakajima:
> Can we make like a regular chat attachment to upload and load seamlessly? 🙇🏼

So we don't recommend attachments get autoloaded by default, since it presents somewhat of a security risk. What do you think of adding the ability to automatically load attachments for "Following" conversations that are under a certain size, maybe 100MB?

> Also is it possible to send text along with attachments?

Not yet, but we can add this soon.

bigint:
> So we don't recommend attachments get autoloaded by default, since it presents somewhat of a security risk. What do you think of adding the ability to automatically load attachments for "Following" conversations that are under a certain size, maybe 100MB?
> 
> >

yea good idea to add only for following users and we have have load button in requested profiles

Also we can explain something like this for users like twitter does instead just showing load button

image

nakajima:
Cool, I'll get goin on that.

bigint:
Amazing you are great @nakajima 💪👏

nakajima:
Ok, just pushed https://github.com/lensterxyz/lenster/pull/2149/commits/528db214b4600b33347a3f9cc4ca4b1098c075e8. This makes it so that we auto load attachments for convos where you're following the other profile (unless the attachment is greater than 100mb).

I also introduced a new cache using indexed db to prevent loading the same attachments multiple times, since it can be a bit expensive.

Some screenshots:

|not following|too big|loader|
|-|-|-|
|Screenshot 2023-03-24 at 1 28 37 PM|Screenshot 2023-03-24 at 1 29 29 PM|Screenshot 2023-03-24 at 1 29 39 PM|

apps/web/src/components/Messages/Composer.tsx
```diff
@@ -1,39 +1,107 @@
...
+  const canSendMessage = !disabledInput && !sending && (message.length > 0 || attachment);
 
   const handleSend = async () => {
     if (!canSendMessage) {
       return;
     }
     setSending(true);
-    const sent = await sendMessage(message);
+
+    var sent: boolean;
```
harish-sethuraman, line 69:
```suggestion
    let sent: boolean;
```

apps/web/src/components/Messages/Composer.tsx
```diff
@@ -1,39 +1,107 @@
...
+      const file = new File([encryptedEncodedContent.payload], 'XMTPEncryptedContent', {
+        type: attachment.mimeType
+      });
+
+      const lensterAttachment = await uploadFileToIPFS(file);
+      const cid = lensterAttachment?.item.replace('ipfs://', '');
+      const url = `https://${cid}.ipfs.w3s.link`;
+
+      const remoteAttachment: RemoteAttachment = {
+        url: url,
```
harish-sethuraman, line 85:
```suggestion
        url,
```

apps/web/src/components/Messages/Composer.tsx
```diff
@@ -58,27 +126,77 @@ const Composer: FC<ComposerProps> = ({ sendMessage, conversationKey, disabledInp
...
+      fileReader.readAsArrayBuffer(file);
+    } else {
+      setAttachment(null);
+    }
+  };
+
+  function onDismiss() {
+    setAttachment(null);
+
+    const el = fileInputRef?.current;
```
harish-sethuraman, line 159:
```suggestion
    const el = fileInputRef.current;
```
all reference will always have `current`?

apps/web/src/components/Messages/MessageContent.tsx
```diff
@@ -0,0 +1,32 @@
...
+type MessageContentProps = {
+  message: DecodedMessage;
+  profile: Profile | undefined;
+  sentByMe: boolean;
+};
+
+const MessageContent = ({ message, profile, sentByMe }: MessageContentProps): JSX.Element => {
+  function content(): JSX.Element {
+    if (message.error) {
+      return <span>{`Error: ${message.error}`}</span>;
```
harish-sethuraman, line 17:
```suggestion
      return <span>Error: {message.error}</span>;
```

bigint:
@nakajima Loader state will cause CLS, can we keep the same width as warn state for loading card 🙇🏼 ?

bigint:
Also build failed 🙇🏼 

image

nakajima:
Pushed a fix for that build error!

> Loader state will cause CLS, can we keep the same width as warn state for loading card

The loader state only appears when you click "View" on an unloaded attachment, so I don't think there should really be more than one on the screen at once. It probably doesn't represent a big CLS issue in that case I think?

bigint:
@nakajima yea, it will still be CLS even after clicking. Here is an example video

Consider we might have some kind of reaction button next to the warning card it will get shifted on clicking view button and there is possibility to react for the message.

[layout-instability2.webm](https://user-images.githubusercontent.com/69431456/228016146-0bb8feb9-71c3-4cf3-b30b-39cf49f41e4e.webm)

bigint:
@nakajima this one

image

bigint:
@nakajima looks cool, we can use small spinner for this

nakajima:
Screenshot 2023-03-28 at 8 16 43 AM

nakajima:
How does that look ^^

bigint:
@nakajima Reduce the padding-y a bit 🙇🏼

nakajima:
Screenshot 2023-03-28 at 8 22 40 AM

^^

bigint:
Perfect

apps/web/src/store/attachment.ts
```diff
@@ -0,0 +1,89 @@
+import { Localstorage } from 'data/storage';
+import { del, get, set } from 'idb-keyval';
```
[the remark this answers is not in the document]
yash-luna: Like @nakajima mentions, storing too many images is not going to be a problem right away

@bigint do you think we can push this update to gauge adoption and then follow up with a PR to evict old data?

bigint:
@nakajima It would be great if you fix the conflicts, I'll merge after that 🚀 🙇🏼

bigint (approved):
**LGTM** :shipit: 

Rad LGTM

saagu335 (approved):
@nakajima Please fix the conflicts and `@` me I'll merge this ASAP 🙇 

PS: we migrated to pnpm from yarn

bigint:
Gonna close this for now!

We need to fix the speed issue first, later we can ship more features 🙇🏼 

RN it takes 40s to load my inbox 😢

bigint:
@nakajima It would be great if you resolve the conflicts 🙇🏼

daria-github:
@bigint conflicts should be resolved in this PR: https://github.com/lensterxyz/lenster/pull/2845

bigint:
Closing this and continued here #2845

PR #2154, opened 2023-03-23
chore: move ui components to ui package

harish-sethuraman:
move more components


bigint:
Gonna review this now 🙇🏼

packages/ui/LightBox.tsx
```diff
@@ -52,10 +50,10 @@ export const LightBox: FC<LightBoxProps> = ({ show, url, onClose }) => {
                   href={url}
                   target="_blank"
                   rel="noopener noreferrer"
-                  onClick={stopEventPropagation}
+                  onClick={(event: SyntheticEvent) => event.stopPropagation()}
```
bigint, line 53: stopEventPropagation still lives in utils you can import here and use it

packages/ui/LightBox.tsx
```diff
@@ -1,7 +1,5 @@
 import { Dialog, Transition } from '@headlessui/react';
```
bigint, line 1: So without installing any deps will it still work?

packages/ui/Input.tsx
```diff
@@ -2,10 +2,9 @@ import clsx from 'clsx';
 import dynamic from 'next/dynamic';
```
bigint, line 2: This too!! not installed, I'm sure if we had pnpm it could have yelled on us, yarn is doing some magic, can we also just try with pnpm to make sure everything is installed properly?

bigint (changes requested):

packages/ui/Input.tsx
bigint: @harish-sethuraman IMO we dont want to install next for this one import, we can use react lazy instead here

harish-sethuraman: We have image import too.

bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #2181, opened 2023-03-25
fix: let NftPicture set the correct chain ID

thoughteer:
## What does this PR do?

The default chain ID for the form should be taken from the profile picture if it's been already configured. If not, we should fallback to Mainnet.

Currently, it always shows me Ethereum, even though my NFT avatar is on Polygon. And that's really confusing.

## Type of change

## How should this be tested?

Unfortunately, there are no tests for the settings page. Hopefully, it's trivial enough to be accepted anyway.

## Checklist


thoughteer:
Damn, apparently, this requires updating `ProfileSettingsDocument` in the lens package as well. Even though the `NftImage` schema supports `chainId`, it's not requested from the Lens API.

thoughteer:
Now it works perfectly.

bigint (approved):
**LGTM** :shipit: 

Rad LGTM

PR #2225, opened 2023-03-27
fix: use ternary operator in clsx instead of object

bigint:


bigint:
Thanks @harish-sethuraman for finding these 🙏🏼

Issue #2258, opened 2023-03-29
Searching with the character "#" in search bar returns 404 for all further searches


GorillaEth:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When using a hashtag '#' anywhere in search menu. 

You get a 404 and all further searches return 404.

Let's say i search for "#BTC". 

The type parameter in the url changes to 'undefined' and stays there for next searches

### Expected Behavior

The search usually returns results and the type parameter is either 'pubs' for publications or 'profiles' for profiles

### Steps To Reproduce

Search with the character '#' and you'll get the problem

### What platform(s) does this occur on?

Web

### What browser(s) does this occur on?

Firefox

PR #2280, opened 2023-03-31
fix: use encodeURIComponent to escape reserved characters in search query

wilsonplau:
## What does this PR do?

Fixes https://github.com/lensterxyz/lenster/issues/2258. The query was not being properly escaped, so certain reserved characters like `#` were causing issues in the search box and in links. Previously, in the Search page, when `#` is appended to the URL as part of the `searchText`, a character like `#` would invalidate rest of the URL following it, so `query.type` would be empty, therefore returning a 404. 

One further note is I wasn't sure what the proper handling for if `query.q` was `undefined` here. Currently, it would lead to a 404 page, which is the same as the existing behaviour. 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## How should this be tested?

- [X] Searching for strings including reserved characters like `#` or `$` using the search box now properly returns search results. 
- [X] Clicking the sidebar (Profiles / Publications) in the Search results page results in proper navigation with the escaped searchText

I haven't updated / added automated tests because I don't think there are any existing tests for the Search feature in general.


bigint (approved):
**LGTM** :shipit: 

Rad LGTM

Issue #2084, opened 2023-03-18
Fix nextjs 13.2.5 build problem


bigint:

THEREALMON:
gm can you tell more details about nextjs 13.2.5 build problem

bigint:
@THEREALMON just bump the version and run yarn build 🙇

THEREALMON:
sure I'll help

bigint:
no one picked this! fell free to pick this @THEREALMON

bigint:
@THEREALMON got any change to find the issue 🙇 ?

THEREALMON:
sorry sir my account was compromised just clearing thing and get back on this.. they taken my lens acct. too 😭

bigint:
It's issue with `@lingui/swc-plugin` https://github.com/lingui/swc-plugin/issues/39

bigint:
Fixed

PR #2292, opened 2023-04-01
fix: update next to latest version (#2084)

en0c-026:


bigint:
@en0c-026 CI got failed 🙇🏼

en0c-026:
> @en0c-026 CI got failed 🙇🏼

I have been investigating and the problem comes from Sentry. I found the following note in their docs:
"While Next.js version 13 is itself supported, the new features it introduces are not yet. In particular, building with Turbopack will currently cause the Sentry SDK not to be included in your app."
So we would have to wait for Sentry to be compatible before updating.

bigint:
Closing this

PR #2757, opened 2023-04-30
Fix Duplicate property name 'import/no-anonymous-default-export' in object literal

SACHINnANYAKKARA:


bigint:
Fixed in #2761 cuz we have in more places

PR #2769, opened 2023-05-01
fix: prioritize nullish coalescing

l2dy:
### <samp>🤖 Generated by Copilot at b4caca8</samp>

Refactored `ActionType` component to simplify and optimize rendering of timeline actions. Removed unnecessary parentheses from `feedItem.comments?.length ?? 0` in `EventType` component.

## Related issues

N/A

## Type of change

### <samp>🤖 Generated by Copilot at b4caca8</samp>

*  Simplify expression for comment count in `ActionType` component ([link](https://github.com/lensterxyz/lenster/pull/2769/files?diff=unified&w=0#diff-2fb7f5a4989580a7152f0c9b3d17b1b58785726241b5308c3c7025f656f735f6L23-R23))

## Emoji

🧹📖🚀

### DIFF
--- a/apps/web/src/components/Home/Timeline/EventType/index.tsx
+++ b/apps/web/src/components/Home/Timeline/EventType/index.tsx
@@ -20,7 +20,7 @@ interface ActionTypeProps {
 const ActionType: FC<ActionTypeProps> = ({ feedItem }) => {
   const publication = feedItem.root;
   const isComment = publication.__typename === 'Comment';
-  const showThread = isComment || (feedItem.comments?.length ?? 0 > 0);
+  const showThread = isComment || (feedItem.comments?.length ?? 0) > 0;
 
   const canCombined = getCanCombined([
     feedItem.mirrors.length,


apps/web/src/components/Home/Timeline/EventType/index.tsx
```diff
@@ -20,7 +20,7 @@ interface ActionTypeProps {
 const ActionType: FC<ActionTypeProps> = ({ feedItem }) => {
   const publication = feedItem.root;
   const isComment = publication.__typename === 'Comment';
-  const showThread = isComment || (feedItem.comments?.length ?? 0 > 0);
+  const showThread = isComment || (feedItem.comments?.length ?? 0) > 0;
```
bigint, line 23: We dont have any ops to do first, so we don't need to follow BODMAS rule here 🙇🏼

l2dy: According to [JavaScript operator precedence](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence), `(feedItem.comments?.length ?? 0 > 0)` is equivalent to `(feedItem.comments?.length ?? (0 > 0))`, which does not return a boolean if `length` exists. For example,

```
> 1 ?? 0 > 0
1
> (1 ?? 0) > 0
true
```

l2dy: If returning a number is expected, the expression could also be simplified into `isComment || (feedItem.comments?.length ?? false)`.

@bigint What do you think?

bigint: Ahh this makes sense 🙏🏼 

Reopened the PR 💟 

Thanks for explaining

PR #2866, opened 2023-05-16
feat: universal inbox

bhavya2611:
## What does this PR do?

This makes the lenster inbox universal

## Type of change

## Explanation of the changes
With this PR, if a user has conversations with addresses, outside lens, they will show up in the other tab on the inbox page, and the user will be able to receive and send messages to these addresses. Important to note, user can still only start conversations with people in lens and not any random wallet address on XMTP, from lenster

## Emoji

 🚀


bigint (changes requested):

apps/web/src/components/Messages/MessageHeader.tsx
```diff
@@ -46,21 +55,43 @@ const MessageHeader: FC<MessageHeaderProps> = ({ profile }) => {
...
-          followSource={FollowSource.DIRECT_MESSAGE_HEADER}
-        />
-      ) : (
-        <Unfollow
-          showText
-          profile={profile}
-          setFollowing={setFollowingWrapped}
-        />
+      {profile && (
+        <div>
```
daria-github, line 79: this can just be a fragment here

apps/web/src/components/utils/hooks/useGetConversation.tsx
```diff
@@ -20,33 +20,37 @@ const useGetConversation = (conversationKey: string, profile?: Profile) => {
...
     }
     if (selectedConversation) {
       setMissingXmtpAuth(false);
       return;
     }
     const createNewConversation = async () => {
-      const conversationId =
-        parseConversationKey(conversationKey)?.conversationId;
-      const canMessage = await Client.canMessage(profile.ownedBy, {
+      const conversationId = conversationKey.split('/')[0];
```
daria-github, line 31: do we need optional chaining here?

apps/web/src/lib/shortAddress.ts
```diff
@@ -0,0 +1,4 @@
+export const shortAddress = (addr: string): string =>
```
daria-github, line 1: a unit test would be good to add for this helper

bhavya2611: Removed this helper

apps/web/src/components/utils/hooks/useGetConversation.tsx
bhavya2611: Because there is no HTML child immediate that the dom can find, es lint throws an error, saying you don't need a fragment here, till there is at least one child

bhavya2611: 👍🏻

apps/web/src/components/Messages/MessageHeader.tsx
bhavya2611: Because there is no HTML child immediate that the dom can find, es lint throws an error, saying you don't need a fragment here, till there is at least one child

bigint:
https://github.com/lensterxyz/lenster/assets/69431456/a7dfb4f0-489d-4b6d-bbe4-f25629b799b5

@bhavya2611 I see on clicking the convo it hides everything on the left 😕

apps/web/src/components/Messages/PreviewList.tsx
```diff
@@ -97,7 +97,7 @@ const PreviewList: FC<PreviewListProps> = ({
     );
     await persistProfile(conversationKey, profile);
     const selectedTab: TabValues = profile.isFollowedByMe
-      ? 'Following'
+      ? 'Lens'
       : 'Requested';
```
bigint, line 101: We can move this string to an enum so we can reuse everywhere without hassle 🙇🏼

bigint (changes requested):

bigint (approved):

bigint:
Thanks for the amazing work @bhavya2611 👏🏼

PR #2882, opened 2023-05-17
feat: send message and attachment

elisealix22:
## What does this PR do?

Allows users to send both an attachment and a message. Also adds fallback text for XMTP apps that don't support attachments yet. Finally, it makes it so you can't dismiss an attachment while it is being sent.

## Related issues

N/A

## Type of change

## Explanation of the changes

Go to send an attachment or a message. Either or both should work.

## Emoji

👯


apps/web/src/components/Messages/Composer.tsx
```diff
@@ -111,8 +114,8 @@ const Composer: FC<ComposerProps> = ({
       );
 
       const uploadedAttachment = await uploadFileToIPFS(file);
-      const cid = sanitizeDStorageUrl(uploadedAttachment?.original.url);
-      const url = `${IPFS_GATEWAY}${cid}`;
+      const hash = uploadedAttachment?.original.url.replace('ipfs://', '');
```
bigint, line 117: Do we need to replace?

elisealix22: Yes there is currently a bug where we are double prefixing which causes an endless spinner when loading remote attachments!

elisealix22: @bigint here's what the url looks like in production currently unfortunately. We didn't catch it at first because cached attachments look great. So only remote attachments are failing with 404 & bad url. 

Screenshot 2023-05-18 at 10 11 54 AM

bigint: I guess while storing you guys push with the gateway 👀 ? no?

Id suggest always push the `ipfs://` thing instead with gateway! cuz gateways are not stable and meant to change 🙇🏼

elisealix22: Ahhh I see your point! I don't the stripping step is necessary since the sanitize also does that and appends the gateway. I just tested without this line and it works. We just need to remove prepending the gateway here and let the sanitize method do that. I updated to account for that!

bigint: amazing

bigint (approved):
**LGTM** :shipit: 

Rad LGTM

bigint:
image

Maybe we can merge text and attachment together! that's for next 🙇🏼

PR #2973, opened 2023-05-30
feat: suggested follows by MadFi

imthatcarlos:
## What does this PR do?

Replaces the component for `Recommended Profiles` in the explore page with `Suggested Follows`
- API request is made to `https://api.madfinance.xyz` (requires api key in header)
- `?profileId` query param is included
- profiles in the response are based on matching ML tags (lens interests)
- if the profile has not been tagged - add them to a queue to be tagged - and return trending profiles
- promoted profiles for a specific tag (lens interest) are shown at the top of the list with a Promoted tag
- when promoted profiles are followed, an impression is logged for lens client rev share

Demo: https://lenstube.xyz/watch/0x21c0-0x67-DA-5bb6c621
API docs: https://docs.madfinance.xyz/api/suggested-follows

## Type of change

## Explanation of the changes

- new component `Suggested Follows` replaces `RecommendedProfiles`
- new package `madfi` for hooks to 1) fetch profiles 2) log impression

## Emoji

💸


bigint:
Hey thanks for the PR, but can we have those in explore page (like you can create section called explore profiles and plugin there), cuz we don't want to disturb core feature on the home page 🙇🏼

imthatcarlos:
@bigint understood! i reverted the change to the home page. for the explore page, im not sure where to place this new section. the left side seems to be a specific feed for posts, leaving only the right side for profile list. what if we left it there?

Screen Shot 2023-05-31 at 6 45 07 PM

bigint:
@imthatcarlos WDYM `creator eco` in the screenshot? will that change?

packages/madfi/utils.ts
bigint, line 1: we should not break forks and other dev envs, being all GET endpoints why we need API key?

bigint (changes requested):

bigint:
PS: CI breaks 😕

packages/madfi/utils.ts
imthatcarlos: yeah so this is where I was hoping for some guidance. we could put this feature behind a feature flag, and maybe it's best that if we _don't_ have an API key, we just see the recommended profiles component.

to answer your question, this GET request can be expensive if the given `profileId` in the query params has **not** been tagged; we spin up a job that runs our ML pipeline to return matching creators on the next uncached request. Finally, the api key is especially important considering the `/impression` POST request - that helps us with the client rev share.

imthatcarlos:
@bigint - I conditionally render the new component if there is an api key in the env. Also, merged in `main` which had a lot of changes since yesterday. Finally, not sure what's going on with the CI, it always says `waiting for status`

bigint:
Closing this, we can revisit this idea later 🙇🏼

Issue #3347, opened 2023-07-24
Automate Fetching and Verification of Lens Profile Followers through GitHub Actions


bigint:
Currently, the process of adding followers from a Lens profile to our verified profile list is manual, involving individual profile ID addition. This method is inefficient and prone to errors. Hence, we aim to automate this process through GitHub Actions.

Objective:
The main goal is to automate the fetching of all followers from a given Lens profile, iterating over these followers, and storing their profile IDs in a file named 'verified.ts'. This will allow us to add these profiles to our verified profile list, bestowing them with a checkmark symbol to signify their verified status.

Specific tasks are as follows:

1. Develop a script/functionality to fetch all followers from a Lens profile.
2. Iterate through each follower and extract the profile ID.
3. Save these profile IDs in a file named 'verified.ts'.
4. Automate the entire process to run every one hour using GitHub Actions.

By doing this, we can significantly enhance our efficiency, reduce manual workload, and improve the accuracy of our verified profile list.

PR #3358, opened 2023-07-24
feat: automated fetching and verification of lens profile

spantheslayer:
## What does this PR do?
Automates Fetching and Verification of Lens Profile from  following

## Related issues
No

Fixes # (issue)
https://github.com/lensterxyz/lenster/issues/3347

## Type of change

## Explanation of the changes
It adds automation which takes lens profile following and adds them to verified list

### DIFF
--- a/packages/data/package.json
+++ b/packages/data/package.json
@@ -15,5 +15,8 @@
     "@lenster/config": "*",
     "@types/node": "^20.4.4",
     "typescript": "^5.1.6"
+  },
+  "dependencies": {
+    "node-fetch": "^3.3.1"
   }
 }
--- a/packages/data/package.json
+++ b/packages/data/package.json
@@ -15,8 +15,5 @@
     "@lenster/config": "*",
     "@types/node": "^20.4.4",
     "typescript": "^5.1.6"
-  },
-  "dependencies": {
-    "node-fetch": "^3.3.1"
   }
 }
--- a/packages/data/verify.js
+++ b/packages/data/verify.js
@@ -26,8 +26,8 @@ async function main() {
 
   let cursor = null;
   while (true) {
-    const fetch = await import('node-fetch');
-    const response = await fetch.default(url, {
+    // const fetch = await import('node-fetch');
+    const response = await fetch(url, {
       method: 'POST',
       headers: {
         'Content-Type': 'application/json'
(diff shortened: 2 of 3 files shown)


packages/data/package.json
```diff
@@ -15,5 +15,8 @@
     "@lenster/config": "*",
     "@types/node": "^20.4.4",
     "typescript": "^5.1.6"
+  },
+  "dependencies": {
+    "node-fetch": "^3.3.1"
```
bigint, line 20: By default there is fetch right, no need of this, try removing and run

packages/data/verify.js
```diff
@@ -0,0 +1,178 @@
...
+
+        // Load the verified.ts file
+        fs.readFile('./verified.ts', 'utf8', (err, tsData) => {
+          if (err) {
+            console.error('Error reading the verified.ts file:', err);
+            return;
+          }
+
+          const regex =
+            /export const mainnetVerified = \[[\S\s]*?...mainnetLensTeamMembers,[\S\s]*?...mainnetStaffs\n];/;
```
bigint, line 87: We don't wanna spread those two arrays, cuz by default we follow all people need to be verified

bigint (changes requested):
Do we really need resp.json instead is it possible to store values on local variable also really we don't want to save the username in comment just array of profile Id is enough

packages/data/package.json
spantheslayer: removed the node-fetch pkg and enabled local variable to store the values

packages/data/verify.js
spantheslayer: I am doing this to append the new data into the verified array before getting the unique data.

either way I have to take it to consideration even if I want to do it with existing data that I get after parsing the response, it's just would have to do the same thing post I get the unique data.

spantheslayer:
@bigint please check, I have updated the code as requested.

packages/data/verify.js
```diff
@@ -0,0 +1,149 @@
...
+        }
+        pageInfo {
+          prev
+          next
+        }
+      }
+    }
+  `;
+
+  const url = 'https://api.lens.dev/';
```
bigint, line 24:
```suggestion
  const url = 'https://api.lens.dev';
```

packages/data/verify.js
```diff
@@ -0,0 +1,149 @@
...
+  const allResponses = [];
+
+  let cursor = null;
+  while (true) {
+    // const fetch = await import('node-fetch');
+    const response = await fetch(url, {
```
bigint, line 30:
```suggestion
  while (true) {
    const response = await fetch(url, {
```

packages/data/verify.js
```diff
@@ -0,0 +1,149 @@
...
+        if (err) {
+          console.error('Error reading the verified.ts file:', err);
+          return;
+        }
+
+        const regex =
+          /export const mainnetVerified = \[[\S\s]*?mainnetLensTeamMembers,[\S\s]*?...mainnetStaffs\n];/;
+        const match = tsData.match(regex);
+        // console.log('match:', match);
+
```
bigint, line 82:
```suggestion
        const match = tsData.match(regex);

```

packages/data/verify.js
```diff
@@ -0,0 +1,149 @@
...
+          const existingProfiles = new Set(
+            match[0]
+              .split('\n')
+              .map((line) => line.trim())
+              .slice(1, -3)
+          );
+
+          // console.log('existingProfiles:', existingProfiles);
+
+          const regex = /'0x([\dA-Fa-f]+)'/g;
```
bigint, line 94:
```suggestion
          );

          const regex = /'0x([\dA-Fa-f]+)'/g;
```

packages/data/verify.js
```diff
@@ -0,0 +1,145 @@
...
+            'mainnetVerified array not found in the verified.ts file.'
+          );
+        }
+      });
+    } catch (error) {
+      console.error('Error:', error);
+    }
+
+    // console.log(`Total number of IDs: ${allResponses.length}`);
+  } else {
```
bigint, line 140:
```suggestion
    }
  } else {
```

packages/data/verify.js
```diff
@@ -0,0 +1,143 @@
...
+      // Load the verified.ts file
+      fs.readFile('./verified.ts', 'utf8', (err, tsData) => {
+        if (err) {
+          console.error('Error reading the verified.ts file:', err);
+          return;
+        }
+
+        const regex =
+          /export const mainnetVerified = \[[\S\s]*?mainnetLensTeamMembers,[\S\s]*?...mainnetStaffs\n];/;
+        const match = tsData.match(regex);
```
bigint, line 79: I'm mean mainnetLensTeamMembers and mainnetStaffs is not needed

PR #3685, opened 2023-09-01
fix: do not show when a profile follows itself

### LINKED ISSUE #3683: Do not show when a profile follows itself
### Summary

Technically, the Lens protocol allows for a profile to follow itself.
Decided that this should not be reflected in Lenster, so any info that a profile follows itself should be hidden.

https://lenster.xyz/posts/0x05-0x1f91-DA-24808777

Screenshot_2023-09-01_10-46-38

foolo:
## What does this PR do?

If the signed-in user, appears in another user's (or in rare cases, their own) Following or Followers list, then hide the follow/unfollow button next to the signed-in user.

**sceenshot**

Screenshot_2023-09-01_18-12-21

### <samp>🤖 Generated by Copilot at 91f0502</samp>

This pull request fixes a bug where the user could see and follow themselves on their own profile page. It modifies the `Followers.tsx` and `Following.tsx` components to exclude the user from the lists and actions.

## Related issues

Fixes #3683 

## Type of change

### <samp>🤖 Generated by Copilot at 91f0502</samp>

*  Filter out the current profile from the lists of followers and followings in `Followers.tsx` and `Following.tsx` to prevent the user from seeing themselves as a follower or following of their own profile ([link](https://github.com/lensterxyz/lenster/pull/3685/files?diff=unified&w=0#diff-55dc73404c164c7379148792a303aad01fcc854056e412dd1a986f59e54a7d58L31-R34), [link](https://github.com/lensterxyz/lenster/pull/3685/files?diff=unified&w=0#diff-26b930a1517cdb6b85350e7d2aec8b151c8bde2181a998c5f71f82cca6e70903L28-R32))
*  Hide the follow and unfollow buttons for the current profile in `Followers.tsx` and `Following.tsx` to prevent the user from following or unfollowing themselves ([link](https://github.com/lensterxyz/lenster/pull/3685/files?diff=unified&w=0#diff-55dc73404c164c7379148792a303aad01fcc854056e412dd1a986f59e54a7d58R100-R102), [link](https://github.com/lensterxyz/lenster/pull/3685/files?diff=unified&w=0#diff-26b930a1517cdb6b85350e7d2aec8b151c8bde2181a998c5f71f82cca6e70903L105-R110))
*  Import and use the `useAppStore` hook in `Following.tsx` to get the current profile from the app state ([link](https://github.com/lensterxyz/lenster/pull/3685/files?diff=unified&w=0#diff-26b930a1517cdb6b85350e7d2aec8b151c8bde2181a998c5f71f82cca6e70903R13), [link](https://github.com/lensterxyz/lenster/pull/3685/files?diff=unified&w=0#diff-26b930a1517cdb6b85350e7d2aec8b151c8bde2181a998c5f71f82cca6e70903R23))

## Emoji

🚫🛠️🗃️

### DIFF
--- a/apps/web/src/components/Profile/Followers.tsx
+++ b/apps/web/src/components/Profile/Followers.tsx
@@ -28,10 +28,7 @@ const Followers: FC<FollowersProps> = ({ profile }) => {
     skip: !profile?.id
   });
 
-  const followers = data?.followers?.items.filter(
-    (follower) => profile?.id !== follower?.wallet?.defaultProfile?.id
-  );
-
+  const followers = data?.followers?.items;
   const pageInfo = data?.followers?.pageInfo;
   const hasMore = pageInfo?.next;
 
--- a/apps/web/src/components/Profile/Followers.tsx
+++ b/apps/web/src/components/Profile/Followers.tsx
@@ -28,7 +28,10 @@ const Followers: FC<FollowersProps> = ({ profile }) => {
     skip: !profile?.id
   });
 
-  const followers = data?.followers?.items;
+  const followers = data?.followers?.items.filter(
+    (follower) => profile?.id !== follower?.wallet?.defaultProfile?.id
+  );
+
   const pageInfo = data?.followers?.pageInfo;
   const hasMore = pageInfo?.next;
 
@@ -92,6 +95,9 @@ const Followers: FC<FollowersProps> = ({ profile }) => {
                   followUnfollowSource={FollowUnfollowSource.FOLLOWERS_MODAL}
                   showBio
                   showFollow={
+                    currentProfile?.id !== follower?.wallet?.defaultProfile?.id
+                  }
+                  showUnfollow={
                     currentProfile?.id !== follower?.wallet?.defaultProfile?.id
                   }
                   showUserPreview={false}
--- a/apps/web/src/components/Profile/Following.tsx
+++ b/apps/web/src/components/Profile/Following.tsx
@@ -10,6 +10,7 @@ import { t, Trans } from '@lingui/macro';
 import { motion } from 'framer-motion';
 import type { FC } from 'react';
 import { Virtuoso } from 'react-virtuoso';
+import { useAppStore } from 'src/store/app';
 
 interface FollowingProps {
   profile: Profile;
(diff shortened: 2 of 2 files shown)


foolo:
@bigint, Updated according to feedback.

apps/web/src/components/Profile/Followers.tsx
```diff
@@ -94,6 +94,9 @@ const Followers: FC<FollowersProps> = ({ profile }) => {
                   showFollow={
                     currentProfile?.id !== follower?.wallet?.defaultProfile?.id
                   }
+                  showUnfollow={
+                    currentProfile?.id !== follower?.wallet?.defaultProfile?.id
+                  }
```
bigint, line 99: I guess we can make a common props for both 🙇🏼

foolo: good point! i'll update it

foolo: @bigint Just want to double check. If we replace showFollow and showUnfollow, with a common prop, (let's say showFollowUnfollow), then it would also change the behavior a bit in other places where UserProfile is used. We would show an Unfollow button in more places.

Example: I am following TeaParty. With this change I would now see an Unfollow in the circle in this screenshot below. Is this desired? I think it would be an improvement.

Screenshot_2023-09-02_11-56-37

bigint: These are for current account only right? how it will hide for other account which is not owned by you 🤔

foolo: A bit hard to explain, but i'll try  🙂
First, yes, this is about when currentProfile sees a another persons profile in in a list or somewhere else at the site.
Also, this is just a small change that I wanted to mention, about the consequence of using a common boolean prop for showFollow and showUnfollow. No big change, but a slight change in behavior, as follows:

If we replace, showFollow and showUnfollow with a common prop, "showFollowUnfollow", when we will **show** the unfollow button in more places, for example in the circle in the screenshot. Because before, in many uses of UserProfile, we only passed showFollow as a prop (and not showUnfollow). But if we replace these with a common boolean, the logic will be the same for follow and unfollow, and thus, we will show **Unfollow** button on all places where we only showed **Follow**.

foolo: Addition: I hope this answers your question. If not, maybe we should instead try to write test cases first, before implementing the logic? Then the expected behavior will be clear from the beginning.

Also one need to keep in mind that there are **three** parties involved in the logic. A: The current account, B: The account who owns the Following/Followers list, and C. The account of each profile displayed in the list. And in each case A, B and C may or may not be the same account.

bigint: Ahh i got you what you mean 🙌🏼  makes sense 🚀

bigint (approved):

PR #3898, opened 2023-10-08
fix: text overlapping issues on mobile browsers

### LINKED ISSUE #3745: [ENG-10] Text is overlapping

### Current Behavior

Text publications with a larger amount of text on Lenster are overlapping on mobile browsers. I've tested Safari, Brave, and Zerion's built-in browsers, but none of them display the textual data without overlapping.

IMG_6206
IMG_6207

### Expected Behavior

Text lines should be displayed one below the other, without being grouped or overlapping.

### Steps To Reproduce

Open Lenster on any of the mentioned mobile browsers, and you should encounter the issue. While I haven't tested it on Chrome or other mobile browsers, I would assume that the behavior is the same.
I'm using an iPhone 13 Pro Max with iOS 16.6.0.

Mobile

Others

### Anything else?

**What browser(s) does this occur on?**
 - Every mobile browser I've tried.

I've noticed that there was a previous report for this issue, but it was marked as resolved. However, the issue has returned. Here's the link to the closed report: https://github.com/lensterxyz/lenster/issues/1922"

<sub>[ENG-10](https://linear.app/heyxyz/issue/ENG-10/text-is-overlapping)</sub>

prateekkalra:
## What does this PR do?

This PR fixes the text overlap issue that occurs in text publications in case of long text in iOS mobile browsers.
This issue occurs because of a bug in Safari browser engine when "line-clamp" is used on wrapper elements with multiple blocks.

React markdown generates multiple blocks inside a wrapper, in case of texts seperated by new lines.
For eg. If there are 10 blocks of texts seperated by new line, then the wrapper "div" would container 10 "p" tags.
In this PR, I have limited the wrapper div to have maximum of 1 block in case of ios mobile browsers(all of them uses safari engine- webkit). So, in that case, the users will see only 1 block of code, and see more button. And if 1 block of code is very large, "line-clamp-5" would work flawlessly on the wrapper element. Since there is only 1 block in wrapper div.

This is the best solution possible. There's no CSS workaround, and the JS workaround would get really messy due to the markdown package.
I'm attaching the before and after screenshots below.
hey-safari-before
hey-safari-after

/claim #3745

## Related issues

Fixes #3745

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### <samp>🤖 Generated by Copilot at ee4c76c</samp>

*  Add conditional logic to truncate publication content on iOS mobile devices ([link](https://github.com/heyxyz/hey/pull/3898/files?diff=unified&w=0#diff-6622353deb6c2dcfc2938623f6b6bc3a65857ee5dd42fe6121d16d43eb4f9f73R193-R199), [link](https://github.com/heyxyz/hey/pull/3898/files?diff=unified&w=0#diff-cf87c96f430fb044b052837b8732058bdc59aba03cba72f7f7c43143622ffae1R48-R54))
  - Import `isIOS` and `isMobile` functions from `react-device-detect` library in `DecryptedPublicationBody.tsx` and `PublicationBody.tsx` ([link](https://github.com/heyxyz/hey/pull/3898/files?diff=unified&w=0#diff-6622353deb6c2dcfc2938623f6b6bc3a65857ee5dd42fe6121d16d43eb4f9f73R43), [link](https://github.com/heyxyz/hey/pull/3898/files?diff=unified&w=0#diff-cf87c96f430fb044b052837b8732058bdc59aba03cba72f7f7c43143622ffae1R19))

## Emoji

🐛📱🔒

### DIFF
--- a/apps/web/src/components/Publication/DecryptedPublicationBody.tsx
+++ b/apps/web/src/components/Publication/DecryptedPublicationBody.tsx
@@ -40,7 +40,7 @@ import Link from 'next/link';
 import { useRouter } from 'next/router';
 import type { FC, ReactNode } from 'react';
 import { useState } from 'react';
-import { isMobile, isSafari } from 'react-device-detect';
+import { isIOS, isMobile } from 'react-device-detect';
 import useContractMetadata from 'src/hooks/useContractMetadata';
 import useEthersWalletClient from 'src/hooks/useEthersWalletClient';
 import { useAppStore } from 'src/store/app';
@@ -190,7 +190,7 @@ const DecryptedPublicationBody: FC<DecryptedPublicationBodyProps> = ({
     setIsDecrypting(false);
   };
 
-  if (isSafari && isMobile && showMore) {
+  if (isIOS && isMobile && showMore) {
     const truncatedContent = content?.split('\n')?.[0];
     if (truncatedContent) {
       setContent(truncatedContent);
--- a/apps/web/src/components/Publication/DecryptedPublicationBody.tsx
+++ b/apps/web/src/components/Publication/DecryptedPublicationBody.tsx
@@ -40,6 +40,7 @@ import Link from 'next/link';
 import { useRouter } from 'next/router';
 import type { FC, ReactNode } from 'react';
 import { useState } from 'react';
+import { isMobile, isSafari } from 'react-device-detect';
 import useContractMetadata from 'src/hooks/useContractMetadata';
 import useEthersWalletClient from 'src/hooks/useEthersWalletClient';
 import { useAppStore } from 'src/store/app';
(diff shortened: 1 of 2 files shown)


apps/web/src/components/Publication/DecryptedPublicationBody.tsx
```diff
@@ -189,6 +190,13 @@ const DecryptedPublicationBody: FC<DecryptedPublicationBodyProps> = ({
     setIsDecrypting(false);
   };
 
+  if (isIOS && isMobile && showMore) {
```
bigint, line 193: isIOS will be mobile at the end right?

bigint: I'd suggest to use some other solution instead of hard detecting things, maybe line-clamp in tailwind?

prateekkalra: 1. isIOS will be true for Mac, iPhone and iPad.
2. Tailwind line-clamp seems to work with iphone browsers, but only in case, when wrapper element has a single div.(As mentioned in PR details above) . How can this very specific issue in iphone browser be handled without detecting the user's device?
PS: This issue occurs in all browsers in iPhone & iPad.  (That's why isIOS & isMobile)
https://github.com/duskload/react-device-detect/blob/master/docs/selectors.md

bigint:
Link T-663

prateekkalra:
@bigint Is there anything else that needs to be done, from my end?

bigint (approved):

PR #3973, opened 2023-11-09
Added Livepeer to known iframe sites

Titan-Node:
## What does this PR do?

Added Livepeer URL to known sites for embedded iframe livestreams.

## Explanation of the changes

[Livepeer](https://github.com/livepeer) needs to be allowed to display an iframe URL of the video content to allow viewing of livestreams and video from the Livepeer network.

While building a Lens Livestreaming Application called [Waves](https://github.com/GageVanK/Waves-Lens) we ran into this issue of cross streaming video to [Hey.xyz](https://github.com/heyxyz)

### DIFF
--- a/packages/workers/oembed/src/helper/meta/generateIframe.ts
+++ b/packages/workers/oembed/src/helper/meta/generateIframe.ts
@@ -4,7 +4,8 @@ const knownSites = [
   'tape.xyz',
   'open.spotify.com',
   'soundcloud.com',
-  'oohlala.xyz'
+  'oohlala.xyz',
+  'lvpr.tv'
 ];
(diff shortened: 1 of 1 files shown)


bigint:
Adding just lvpr will not work, it should have proper iframe support 🙇🏼 

Closing this now!

0xcadams:
> Adding just lvpr will not work, it should have proper iframe support 🙇🏼
> 
> Closing this now!

Hey @bigint - could you explain the rationale behind this? We do not set `X-Frame-Options` on our response, so it supports iframing as-is. We can add headers that you'd like to see, just let us know what you need.

PR #4184, opened 2023-12-11
fix: tons of errors using biome

bigint:
## Related issues

Fixes # (issue)

## Type of change

### DIFF
--- a/apps/api/src/utils/feeds/providers/k3l/k3lFeed.ts
+++ b/apps/api/src/utils/feeds/providers/k3l/k3lFeed.ts
@@ -14,7 +14,7 @@ const k3lFeed = async (
     return await k3lPersonalFeed(strategy, profile, limit, offset);
   }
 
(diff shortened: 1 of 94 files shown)


biome.json
```diff
@@ -0,0 +1,36 @@
+{
+	"$schema": "https://biomejs.dev/schemas/1.4.1/schema.json",
+	"organizeImports": {
+		"enabled": false
+	},
+	"files": {
+		"maxSize": 100000000,
+		"ignore": [
+			"dist/*",
```
harish-sethuraman, line 9: rather than doing this you could have run it once on all files and then you could have configured biome to run only on source dir? and other required directories?

bigint: makes sense!

bigint: being monorepo its hard to whitelist files rather than ignore right? no?

harish-sethuraman: true

PR #4259, opened 2023-12-18
feat: redesign select dropdown

### LINKED ISSUE #4246: UI: Design select menu
https://tailwindui.com/components/application-ui/forms/select-menus

Props:

- options[]
  - title: ReactNode 
  - description: ReactNode
  - key: string
  - value: string
-  `...rest` of HTML select

RutamBhagat:
## What does this PR do?
Update Select.tsx UI
/claim #4246

## Related issues
Fixes #4246

## Type of change

## Explanation of the changes
Updated Select.tsx as requested.
Had to make changes to the props since pre-commit hooks were preventing from making a commit, please let me know if I need to make any other changes


RutamBhagat:
https://github.com/heyxyz/hey/assets/72187009/7fb81ce7-1577-48a1-9d12-305b6a2f5f72

Here is a demo

bigint:
Can you show demo with description as well, also trying all props like disabled, selected etc

also looks like spacing is very huge between two items!

RutamBhagat:
and for disabled options I have disabled the active state when hovered also disabled options cant be selected by default or when performing onClick on options

https://github.com/heyxyz/hey/assets/72187009/d06ed050-6851-4700-bac5-9a9d01fd4f3a

bigint:
make sure to update lock file

apps/web/package.json
```diff
@@ -65,6 +65,7 @@
     "remark-breaks": "^4.0.0",
     "remark-linkify-regex": "^1.2.1",
     "shared-zustand": "^2.0.0",
+    "solid": "link:@heroicons/react/20/solid",
```
bigint, line 68: ohh no what is this? 😅

RutamBhagat: Do you want me to replace them with plain svg tags? They are the icons that are used in the select and selected option
image
image

bigint: we already have heroicon installed, you can use it!

RutamBhagat: done

bigint:
Screenshot 2023-12-19 at 12 05 54 AM

also the input ui and select ui should look same!

bigint:
while hover the mouse should use cursor-pointer

bigint:
Screenshot 2023-12-19 at 12 06 43 AM

use shadow-sm and radius xl

bigint:
Screenshot 2023-12-19 at 12 28 40 AM

are we sure both have same shadow?

RutamBhagat:
Updated height of select component and changed shadow and corner radius of options menu 

https://github.com/heyxyz/hey/assets/72187009/bc5b9339-9dc0-4a0b-a837-7daf4d1f0fb1

bigint:
https://github.com/heyxyz/hey/assets/69431456/7e0d1575-87a7-4418-8a20-198141ecfde1

i dont see which one is selected

bigint:
Screenshot 2023-12-19 at 12 29 52 AM

seeing dupe label

RutamBhagat:
> Screenshot 2023-12-19 at 12 28 40 AM
> 
> are we sure both have same shadow?

should I remove shadow from the select button?

bigint:
Screenshot 2023-12-19 at 12 30 06 AM

fonts should be normal, not sm/xs

packages/ui/src/Select.tsx
```diff
@@ -1,43 +1,143 @@
...
 
+type Option = {
+  description?: ReactNode;
+  disabled?: boolean;
+  key?: string;
+  label?: string;
+  selected?: boolean;
+  title?: ReactNode;
+  value?: string;
+};
```
bigint, line 17:
```suggestion
};

```

bigint:
> should I remove shadow from the select button?

yes

bigint:
it should be same as input!

bigint (changes requested):

RutamBhagat:
> Screenshot 2023-12-19 at 12 30 06 AM
> 
> fonts should be normal, not sm/xs

image
image
There is no font-sm or font-xs there are only font-normal, font-medium, and font-semibold to highlight selected items

RutamBhagat:
> Screenshot 2023-12-19 at 12 29 52 AM
> 
> seeing dupe label

fixed this, there was a placeholder

RutamBhagat:
> Screen.Recording.2023-12-19.at.12.29.07.AM.mov
> 
> i dont see which one is selected

Not sure why this is happening, working on it

RutamBhagat:
https://github.com/heyxyz/hey/assets/72187009/45780f88-139c-4150-9b6a-e69e0c6e5d63

@bigint in the Select component i implemented the selected element is shown as active element (It only requires one click) but in the project where the Select components are implemented take double consecutive clicks to change state as shown in the screen recording I am not sure why this is happening can you please suggest me some solution?

bigint:
> @bigint in the Select component i implemented the selected element is shown as active element (It only requires one click) but in the project where the Select components are implemented take double consecutive clicks to change state as shown in the screen recording I am not sure why this is happening can you please suggest me some solution?

why we need to use state here? instead select has own props which has selected or not right?

bigint:
How is tailwind css example looks like?

RutamBhagat:
https://github.com/heyxyz/hey/assets/72187009/84d7a3f0-14b5-4234-beb2-e54f97a5ecb5

@bigint I think I found the issue, this only happens when the options array is placed inside the RFC, when we place the options array outside the RFC the select tag works as expected

RutamBhagat:
> How is tailwind css example looks like?

Even in their tailwindcss React example they have defined the people array outside the RFC
image

RutamBhagat:
https://github.com/heyxyz/hey/assets/72187009/8d034a9a-bf57-4469-9d06-438a8797e136

The Reason.tsx component is working as expected after moving options out of RFC, but in order to do that I had to set every first option as selected: true and rest as false, Should i do this for rest of the components that use Select components?

RutamBhagat:
> Screen.Recording.2023-12-19.at.12.29.07.AM.mov
> 
> i dont see which one is selected

another alternative to solve this is to move the options array into useState 

https://github.com/heyxyz/hey/assets/72187009/9eda8c4a-cf96-4b47-a5d7-388a211e340f

apps/web/src/components/Composer/Actions/CollectSettings/AmountConfig.tsx
```diff
@@ -17,6 +17,12 @@ const AmountConfig: FC<AmountConfigProps> = ({
   allowedTokens,
   setCollectType
 }) => {
+  const [options] = useState(
+    allowedTokens?.map((token) => ({
+      label: token.name,
+      value: token.contractAddress
+    }))
+  );
```
bigint, line 25: its not a good idea to have state everywhere 😕

RutamBhagat: It causes problems with selection otherwise, the only other option is to move options outside of RFC but that causes problems where the option.selected are dependent on other props

bigint:
We need to make this more generic and abstract things on the parent component same like other UI lib does, we can have state inside UI comp instead of all comp using select!

RutamBhagat:
> We need to make this more generic and abstract things on the parent component same like other UI lib does, we can have state inside UI comp instead of all comp using select!

image
yeah it's working after setting new useState inside the Select.tsx with props options as default parameter

bigint:
Screenshot 2023-12-20 at 11 44 10 AM

padding is off compared to input

RutamBhagat:
image
this is with py-[10px] should i decrease it?

bigint:
Screenshot 2023-12-22 at 2 34 23 PM

there is no empty state as well 🙇🏼

RutamBhagat:
image

what should I provide as default state when no options are provided or are undefined?

RutamBhagat:
> We need to make this more generic and abstract things on the parent component same like other UI lib does, we can have state inside UI comp instead of all comp using select!

I have noticed some state problems

https://github.com/heyxyz/hey/assets/72187009/05ac8d61-c0e1-49dd-b2eb-af4a9372075f

its happening because of we moved state inside UI.
when type select is changed the reason sub select options wont change.
I think my previous approach of moving options in the parent component inside useState is the only possible fix for this issue

bigint:
https://headlessui.com/react/combobox

bigint:
https://headlessui.com/react/listbox

RutamBhagat:
> https://headlessui.com/react/listbox

https://github.com/heyxyz/hey/assets/72187009/486a590c-dc67-4b71-b90c-83714a3283bd


I have implemented the components you provided but the selected components are still not showing as active Its exact same issue as earlier. We have to move the options to useState in the parent component or keep it outside of RFC. Passing options directly is causing the selected item being not shown as active and setting useState inside Select.tsx doesn't works either as mentioned earlier

RutamBhagat:
> > https://headlessui.com/react/listbox
> 
> simplescreenrecorder-2023-12-23_13.26.00.mp4
> 
> I have implemented the components you provided but the selected components are still not showing as active Its exact same issue as earlier. We have to move the options to useState in the parent component or keep it outside of RFC. Passing options directly is causing the selected item being not shown as active and setting useState inside Select.tsx doesn't works either as mentioned earlier

@bigint should I proceed with moving options in useState in the parent components?

PR #4347, opened 2023-12-25
feat: migrate headless ui components to radix primitives

### LINKED ISSUE #4327: Migrate away from all headless ui usage to use radix primitives

sdthakral33:
## What does this PR do?
This PR contains the changes for migrating headlessui library to radix ui primitve
## Related issues
/claim #4327
Fixes #4327
## Type of change

## Explanation of the changes
All the components like menus and modals using headlessui components have been migrated to radix ui primitive

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -33,7 +33,7 @@
     "dotenv": "^16.0.0",
     "express": "^4.17.1",
     "express-file-routing": "^3.0.3",
-    "fast-xml-parser": "^4.3.2",
+    "fast-xml-parser": "^4.3.3",
     "linkedom": "^0.16.6",
     "node-zendesk": "^5.0.9",
     "react": "^18.2.0",
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -8,7 +8,7 @@
     "eslint": "^8.56.0",
     "eslint-config-next": "^14.0.4",
     "eslint-plugin-perfectionist": "^2.5.0",
-    "eslint-plugin-prettier": "^5.1.1",
+    "eslint-plugin-prettier": "^5.1.3",
     "eslint-plugin-react": "^7.33.2",
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-unicorn": "^50.0.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -66,8 +66,8 @@ importers:
         specifier: ^3.0.3
         version: 3.0.3(express@4.18.2)
       fast-xml-parser:
-        specifier: ^4.3.2
-        version: 4.3.2
+        specifier: ^4.3.3
+        version: 4.3.3
       linkedom:
         specifier: ^0.16.6
         version: 0.16.6
@@ -432,8 +432,8 @@ importers:
         specifier: ^2.5.0
         version: 2.5.0(eslint@8.56.0)(typescript@5.3.3)
       eslint-plugin-prettier:
-        specifier: ^5.1.1
-        version: 5.1.2(eslint@8.56.0)(prettier@3.1.1)
+        specifier: ^5.1.3
+        version: 5.1.3(eslint@8.56.0)(prettier@3.1.1)
       eslint-plugin-react:
         specifier: ^7.33.2
         version: 7.33.2(eslint@8.56.0)
@@ -6271,7 +6271,7 @@ packages:
       '@react-native-community/cli-tools': 12.3.0
       chalk: 4.1.2
       execa: 5.1.1
-      fast-xml-parser: 4.3.2
+      fast-xml-parser: 4.3.3
       glob: 7.2.3
       logkitty: 0.7.1
     transitivePeerDependencies:
@@ -6284,7 +6284,7 @@ packages:
       '@react-native-community/cli-tools': 12.3.0
       chalk: 4.1.2
       execa: 5.1.1
-      fast-xml-parser: 4.3.2
+      fast-xml-parser: 4.3.3
       glob: 7.2.3
       ora: 5.4.1
     transitivePeerDependencies:
@@ -6668,7 +6668,7 @@ packages:
   /@safe-global/safe-apps-sdk@8.1.0(typescript@5.3.3)(zod@3.22.4):
     resolution: {integrity: sha512-XJbEPuaVc7b9n23MqlF6c+ToYIS3f7P2Sel8f3cSBQ9WORE4xrSuvhMpK9fDSFqJ7by/brc+rmJR/5HViRr0/w==}
     dependencies:
-      '@safe-global/safe-gateway-typescript-sdk': 3.13.3
+      '@safe-global/safe-gateway-typescript-sdk': 3.14.0
       viem: 1.21.4(typescript@5.3.3)(zod@3.22.4)
     transitivePeerDependencies:
       - bufferutil
@@ -6677,8 +6677,8 @@ packages:
       - zod
     dev: false
 
-  /@safe-global/safe-gateway-typescript-sdk@3.13.3:
-    resolution: {integrity: sha512-qBDM469cVCedpBpeTSn+k5FUr9+rq5bMTflp/mKd7h35uafcexvOR/PHZn2qftqV8b1kc9b8t22cPRJ2365jew==}
+  /@safe-global/safe-gateway-typescript-sdk@3.14.0:
+    resolution: {integrity: sha512-/dqU66RvHw50n+7x3nwnJedq8V6iLQyoWitNdjx5cFTBmae+rpP+LvHq+LqZfXJVkB1qNytMdjFjdyES0t79gQ==}
     engines: {node: '>=16'}
     dev: false
 
@@ -10312,7 +10312,7 @@ packages:
       end-of-stream: 1.4.4
       inherits: 2.0.4
       readable-stream: 3.6.2
-      stream-shift: 1.0.1
+      stream-shift: 1.0.2
     dev: false
 
   /eastasianwidth@0.2.0:
@@ -10777,8 +10777,8 @@ packages:
       - typescript
     dev: true
 
-  /eslint-plugin-prettier@5.1.2(eslint@8.56.0)(prettier@3.1.1):
-    resolution: {integrity: sha512-dhlpWc9vOwohcWmClFcA+HjlvUpuyynYs0Rf+L/P6/0iQE6vlHW9l5bkfzN62/Stm9fbq8ku46qzde76T1xlSg==}
+  /eslint-plugin-prettier@5.1.3(eslint@8.56.0)(prettier@3.1.1):
+    resolution: {integrity: sha512-C9GCVAs4Eq7ZC/XFQHITLiHJxQngdtraXaM+LoUFoFp/lHNl2Zn8f3WQbe9HvTBBQ9YnKFB0/2Ajdqwo5D1EAw==}
     engines: {node: ^14.18.0 || >=16.0.0}
     peerDependencies:
       '@types/eslint': '>=8.0.0'
@@ -11258,8 +11258,8 @@ packages:
       strnum: 1.0.5
     dev: false
 
-  /fast-xml-parser@4.3.2:
-    resolution: {integrity: sha512-rmrXUXwbJedoXkStenj1kkljNF7ugn5ZjR9FJcwmCfcCbtOMDghPajbc+Tck6vE6F5XsDmx+Pr2le9fw8+pXBg==}
+  /fast-xml-parser@4.3.3:
+    resolution: {integrity: sha512-coV/D1MhrShMvU6D0I+VAK3umz6hUaxxhL0yp/9RjfiYUfAv14rDhGQL+PLForhMdr0wq3PiV07WtkkNjJjNHg==}
     hasBin: true
     dependencies:
       strnum: 1.0.5
@@ -16413,8 +16413,8 @@ packages:
       duplexer: 0.1.2
     dev: true
 
-  /stream-shift@1.0.1:
-    resolution: {integrity: sha512-AiisoFqQ0vbGcZgQPY1cdP2I76glaVA/RauYR4G4thNFgkTqr90yXTo4LYX60Jl+sIlPNHHdGSwo01AvbKUSVQ==}
+  /stream-shift@1.0.2:
+    resolution: {integrity: sha512-rV4Bovi9xx0BFzOb/X0B2GqoIjvqPCttZdu0Wgtx2Dxkj7ETyWl9gmqJ4EutWRLvtZWm8dxE+InQZX1IryZn/w==}
     dev: false
 
   /stream-transform@2.1.3:


sdthakral33:
Hey @bigint Please review this PR.
Thanks!

bigint:
https://github.com/heyxyz/hey/assets/69431456/5988c143-b3d0-49b6-9ca2-a669fc0a3435

can't close modal on clicking outside

bigint:
https://github.com/heyxyz/hey/assets/69431456/5304a5f3-a3be-4832-8cc4-1d150cbe7e82

scroll is enabled even after opening the modal

bigint:
https://github.com/heyxyz/hey/assets/69431456/b3578a2d-0716-4e6c-9ea9-8f1d32ab17d4

dropdown misaligned

bigint:
https://github.com/heyxyz/hey/assets/69431456/dca25314-2167-4c5a-a16b-67ab134884b8

here too

bigint:
Screenshot 2023-12-26 at 9 32 26 PM

misaligned here too

sdthakral33:
@bigint Thanks for the comments! Working on fixing those.

apps/web/src/components/Composer/Actions/Attachment.tsx
```diff
@@ -13,6 +11,7 @@ import {
   MediaAudioMimeType,
   MediaImageMimeType
 } from '@lens-protocol/metadata';
+import * as DropdownMenuPrimitive from '@radix-ui/react-dropdown-menu';
```
bigint, line 14: The alias name is huge imo

apps/web/src/components/Composer/Actions/Attachment.tsx
```diff
@@ -91,99 +90,118 @@ const Attachment: FC = () => {
     }
   };
 
+  const preventDefaultEvent = (e: Event) => {
+    e.preventDefault();
+  };
```
bigint, line 95: we have a global util called `stopEventPropagation.ts`

apps/web/src/components/Composer/Actions/Attachment.tsx
```diff
@@ -91,99 +90,118 @@ const Attachment: FC = () => {
...
-      </Menu.Button>
-      <MenuTransition show={showMenu}>
-        <Menu.Items
-          className="absolute z-[5] mt-2 rounded-xl border bg-white py-1 shadow-sm focus:outline-none dark:border-gray-700 dark:bg-gray-900"
+    <DropdownMenuPrimitive.Root
+      modal={false}
+      onOpenChange={setShowMenu}
+      open={showMenu}
+    >
+      <div>
```
bigint, line 103: do we need this extra div?

apps/web/src/components/Composer/Actions/Attachment.tsx
```diff
@@ -91,99 +90,118 @@ const Attachment: FC = () => {
...
+                onChange={handleAttachment}
+                type="file"
+              />
+            </label>
+          </DropdownMenuPrimitive.Item>
+          <DropdownMenuPrimitive.Item
+            className="rounded-lg menu-item data-[highlighted]:bg-gray-100 dark:data-[highlighted]:bg-gray-800 data-[disabled]:cursor-default"
             disabled={Boolean(attachments.length)}
-            htmlFor={`audio_${id}`}
+            onSelect={preventDefaultEvent}
```
bigint, line 180: why we need to prevent it?

packages/ui/src/Modal.tsx
```diff
@@ -24,34 +23,12 @@ export const Modal: FC<ModalProps> = ({
...
-            enter="ease-out duration-100"
-            enterFrom="opacity-0 translate-y-4 sm:translate-y-0 sm:scale-95"
-            enterTo="opacity-100 translate-y-0 sm:scale-100"
-            leave="ease-in duration-100"
-            leaveFrom="opacity-100 translate-y-0 sm:scale-100"
-            leaveTo="opacity-0 translate-y-4 sm:translate-y-0 sm:scale-95"
-          >
+    <Dialog.Root open={show}>
+      <Dialog.Portal>
+        <div className="flex min-h-screen items-center justify-center p-4 text-center sm:block sm:p-0 fixed inset-0 z-10 overflow-y-auto">
```
bigint, line 28: is this div not suitable on any other of radix comps?

apps/web/src/components/Explore/index.tsx
```diff
@@ -45,49 +44,43 @@ const Explore: NextPage = () => {
...
-              </Tab.Panel>
-            ))}
-          </Tab.Panels>
-        </Tab.Group>
+          </Tabs.List>
+          <div className={'mt-4'}>
+            <FeedFocusType focus={focus} setFocus={setFocus} />
+          </div>
+          {tabs.map((tab, index) => (
+            <Tabs.Content className={'mt-4'} key={tab.type} value={`${index}`}>
```
bigint, line 79: why we need mt-4 also it should not wrapped under `{}`

apps/web/src/components/Explore/index.tsx
```diff
@@ -45,49 +44,43 @@ const Explore: NextPage = () => {
...
-          <Tab.Panels>
-            {tabs.map((tab) => (
-              <Tab.Panel key={tab.type}>
-                <Feed feedType={tab.type} focus={focus} />
-              </Tab.Panel>
-            ))}
-          </Tab.Panels>
-        </Tab.Group>
+          </Tabs.List>
+          <div className={'mt-4'}>
```
bigint, line 75: why we need mt-4 also it should not wrapped under `{}`

bigint (changes requested):

apps/web/src/components/Composer/Actions/Attachment.tsx
sdthakral33: Yes, This is needed to align the Picture Icon with other icons present

apps/web/src/components/Composer/Actions/Attachment.tsx
sdthakral33: DropdownMenuPrimitive.Item has default handler attached to it because of which the call was not going to its children i.e.  <label>,  to supress this behaviour, used prevent default.

apps/web/src/components/Explore/index.tsx
sdthakral33: Tabs.Content and Tabs-Content in radix ui doesn't have a default margin set. That's why added margin to keep the look of UI intact with the previous one.

apps/web/src/components/Explore/index.tsx
sdthakral33: Tabs.Content and Tabs-Content in radix ui doesn't have a default margin set. That's why added margin to keep the look of UI intact with the previous one.

packages/ui/src/Modal.tsx
sdthakral33: No, it is not needed on other radix components like Menu.
Modal is a generic component called at multiple places.

apps/web/src/components/Composer/Actions/Attachment.tsx
sdthakral33: Sure, will use that.

apps/web/src/components/Composer/Actions/Attachment.tsx
sdthakral33: Will change it from DropMenuPrimitive to DropdownMenu

apps/web/src/components/Composer/Actions/Attachment.tsx
sdthakral33: This `stopEventPropagation.ts` utility cannot be used as I don't want to stop the event propagation, instead of that I want to prevent the default event itself.

sdthakral33:
@bigint Addressed All the review comments and made changes accordingly. Please review the changes again.
Thanks!

bigint:
Screenshot 2023-12-27 at 2 16 42 PM

there is no spacing below the navbar weird 🤔

bigint:
@sdthakral33 can you please make it pixel perfect checking side by side with prod and your branch?

sdthakral33:
> Visit Preview

@bigint I am not to seeing this issue on my machine. I did a side by side check also, it looks fine.
Screenshot 2023-12-27 at 3 08 42 PM (2)

bigint:
It's happening for me when staff mode is enabled, there will be another navbar above the navbar, it works fine on main branch

sdthakral33:
@bignit Is there some way for me to test the behaviour in staff mode?

bigint:
you can just set true on staff mode boolean manual on the code

https://github.com/heyxyz/hey/blob/351a5b4b25c3a23e6fb64b7ccfc14ca5ce306c17/apps/web/src/components/Shared/Navbar/index.tsx#L73

sdthakral33:
@bigint I have fixed that navbar issue, can you pls check it again

apps/web/src/components/Common/Layout.tsx
```diff
@@ -104,9 +106,11 @@ const Layout: FC<LayoutProps> = ({ children }) => {
       <GlobalAlerts />
       <div className="flex min-h-screen flex-col pb-14 md:pb-0">
         <Navbar />
-        <GlobalBanners />
-        <BottomNavigation />
-        {children}
+        <div className={cn({ 'pt-[5.8rem]': staffMode, 'pt-16': !staffMode })}>
```
bigint, line 109: this will create huge huge huge problem, it wraps 100s of 1000s of DOM in this div and an good approach 🙇🏼 what caused this issue?

bigint (changes requested):

apps/web/src/components/Common/Layout.tsx
sdthakral33: We are adding a class to the div, it will not cause any types of re renders in IMO

apps/web/package.json
```diff
@@ -16,9 +16,10 @@
   },
   "dependencies": {
     "@apollo/client": "^3.8.8",
+    "@aws-sdk/client-s3": "^3.478.0",
+    "@aws-sdk/lib-storage": "^3.478.0",
     "@aws-sdk/client-s3": "^3.481.0",
     "@aws-sdk/lib-storage": "^3.481.0",
```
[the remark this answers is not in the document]
sdthakral33: Will remove it!

apps/web/src/components/Common/Layout.tsx
bigint: no rerenders, it adds extra dom to the root!

bigint: https://www.commercegurus.com/dom-size/

sdthakral33: Yeah, I understand but how it can cause a huge huge problem?

bigint: its a new div!

sdthakral33: @bigint Can we discuss it? I can explain you why that div is needed.

sdthakral33: @bigint This div is needed to create a separation between navbar and actual content of the page. It will increase the overall DOM size by just 1. I don't think that it will have any impact on the performance of the application.
There is no other way to handle this. The behaviour of radix components and headless components are quite different. May be we can discuss on discord, if you want more explaination.

bigint: But question is why its affecting areas that are not using radix?

apps/web/package.json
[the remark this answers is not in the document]
sdthakral33: Removed Duplicate Libs!

apps/web/src/components/Common/Layout.tsx
sdthakral33: @bigint Whenever a modal is opened in radix, we will need to make the document element overflow property to hidden to disable the parent scroll, because of which the navbar having sticky position will not work and will not be visible (if parent overflow is hidden, child descendents with sticky position will not work). that's why I have change the navbar position to fixed. Now the problem is that since the position of navbar is fixed, it will hide the content behind it, to solve that we will need to have a wrapper around the content and will need to add padding top to it which equals to the navbar height to make the content visible below navbar.

bigint:
https://github.com/heyxyz/hey/assets/69431456/2eed09ae-d318-445b-8d19-0e48c087c87c

closing is not smooth!

sdthakral33:
> Screen.Recording.2023-12-27.at.10.11.56.PM.mov 
> closing is not smooth!

Fixed this in the latest commit.

apps/web/src/components/Common/Layout.tsx
bigint: where is the code affecting this?

sdthakral33: @bigint Please refer this useEffect.
https://github.com/heyxyz/hey/blob/58cf03abcc89265c7e3693e44b91c20331d4807f/packages/ui/src/Modal.tsx#L27

sdthakral33: Because of which the navbar sticky position will not work and it will be hided from the background, that's why changed navbar position to fixed. Refer
https://github.com/heyxyz/hey/blob/1aa3372b45ee28888ee6de03aae6073229c6e2a5/apps/web/src/components/Shared/Navbar/index.tsx#L72

sdthakral33: @bigint Can you pls help me to understand what you are worrying about in this change?

sdthakral33: @bigint Any Updates?

sdthakral33:
@bigint Any update on this PR?

apps/web/src/components/Common/Layout.tsx
bigint: @sdthakral33 i mean which primitive makes the UI looks ugly?

bigint: im worried because adding a root level dom will increase the dom depth, which affects web vitals drastically

bigint: image

it will add a another parent to the dom, i could have accepted if the dom is appended anywhere in the middle, but parent is huge!

bigint: https://github.com/heyxyz/hey/assets/69431456/f992172f-61a0-4cd9-a699-2bafdf2a5b98

sdthakral33: @bigint If we want to use radix primitive then there is not other way, it will make the body overflow ahidden (part of radix framework), because of which we cannot use sticky navbar with radix. If we will be going to change the position of navbar to fixed, then we will need to add wrapper around the main content just after the navbar to adjust the content according to the fixed navbar.

sdthakral33: > image
> 
> it will add a another parent to the dom, i could have accepted if the dom is appended anywhere in the middle, but parent is huge!

Yeah, I understand that but it will increase the depth by just one. IMO that should not be a problem.

sdthakral33: If that div would have increased the depth by some n (large) number, then there can be web vitals issues.

sdthakral33: @bigint I can fix this. Just gave a thought about it.

sdthakral33: @bigint Fixed this in the latest commit. Please review!

sdthakral33: https://github.com/heyxyz/hey/pull/4347/files#diff-0e16eaa722dcbe43f850552755cc5fafac704a32a39761450c6d2edf5a91a637

sdthakral33: @bigint Any Updates? Is this fine?

bigint: > Radix Primitive Dialog will make the UI look ugly.

Are they removing the style while without even consuming them? I don't see any of the platform using radix doing this to manually add multiple padding, this looks like a antipattern and not the fix itself 🙇🏼

sdthakral33: Screenshot 2023-12-30 at 1 09 10 PM (2)
@bigint Radix UI itself uses padding top to create separation between header and main content. Please refer the above screenshot you can see `padding-top: var(--header-height)`

bigint: but we already have that padding in grid layouts

sdthakral33: @bigint No, we don't have that padding in grid layouts
I think I should add that padding in grid layout.

sdthakral33: @bigint We don't have that padding in grid layouts

sdthakral33: @bigint The “sticky” value of the navbar position property is a mixture of the relative and fixed positioning (which adjusts the content by itself, the actual content will start after the navbar div/header). But with radix, sticky navbar doesn't work since it makes body overflow as hidden (, that's why if we will have to make navbar position to fixed. If we have made navbar position to fixed, then we will need to adjust the content by adding a padding top which creates that separation)
The padding in Grid Layouts is given by considering that the content will just after the Navbar (assuming its sticky position). But since we have changed the position from sticky to fixed, we will need to add this additional padding top.

sdthakral33: Screenshot 2023-12-30 at 1 58 30 PM (2)
Please refer this screenshot and https://michaelmovsesov.com/articles/fix-css-position-sticky-not-working

sdthakral33: @bigint Got the actual issue, which is causing the behaviour of radix modal to deviate from its original behaviour.
https://github.com/heyxyz/hey/blame/dc27d01f3b899cf759279ca39c4bbb3bf1ecd4f5/apps/web/src/styles.css#L7
This line is causing issue, is there a need to apply overflow scroll y on whole html document, I don't think so because we have already used overflow tailwind classes to handle this and anyways applying classes directly to html element is also not good practice.
Do you have some opinions on it?
Will it be fine I remove this?

sdthakral33: @bigint Fixed this in latest commit!

sdthakral33: @bigint Any updates on this? Is now everything fine?

sdthakral33:
@bigint Any updates on this PR? I have made all the changes wrt the review comments and everything now seems to work fine. Can you please review and provide any further comments if you have.
Thanks!

apps/web/src/components/Common/Layout.tsx
sdthakral33: @bigint Any Updates???

sdthakral33:
@bigint Any Updates?

bigint:
I see lot of issues still

image

seeing outer ring on hover

sdthakral33:
> I see lot of issues still
> 
> image
> seeing outer ring on hover

@bigint It is already existing behaviour. It is coming on :focus-visible not on :hover, adding an blue outline. I have not added anything extra. Please see the attached screenshot for reference on testnet.

sdthakral33:
> I see lot of issues still
> 
> image
> seeing outer ring on hover

@bigint Is there any issue other than this one that you have observed?

sdthakral33:
> > I see lot of issues still
> > image
> > seeing outer ring on hover
> 
> @bigint It is already existing behaviour. It is coming on :focus-visible not on :hover, adding a blue outline. I have not added anything extra. Please see the attached screenshot for reference on testnet. Screenshot 2024-01-04 at 12 34 24 PM (2) Screenshot 2024-01-04 at 12 43 12 PM (2)

@bigint Let me know if you want me to fix this behaviour.

bigint:
@sdthakral33 it can happen when using keyboard but when just using the mouse its happening 🙇🏼

bigint:
https://github.com/heyxyz/hey/assets/69431456/fef4dc42-33bf-4740-919c-0e26e0429f5c

sdthakral33:
> @sdthakral33 it can happen when using keyboard but when just using the mouse its happening 🙇🏼

@bigint It is not happening with keyboard on testnet. Please see this video. I can also have the same behaviour by making outline as none. Please let me know your opinion on that.

https://github.com/heyxyz/hey/assets/68169950/4cb8de82-9c62-45ca-acc5-4c12173f73ee

Issue #4240, opened 2023-12-16
Integrate realtime DMs with Push Protocol


bigint:
- Able to chat between two lens profiles (not wallets)
- Should be realtime
- Ability to upload images and videos
- Ability to react on messages
- Ability to reply to messages
- Make sure to follow Hey coding standards

Bounties will be paid on every big milestone as Algora tip

https://push.org/docs/chat/build/

Can jump on a call for a discussion

PS: no one will be assigned, I will review all PRs, lets see who gonna grab em all 😉 
```[tasklist]
### Tasks
- [ ] Add optimisting send in DMs aka fake sending message, same like whatsapp does!
```

b4s36t4:
@bigint I'm taking this up.

b4s36t4:
Was there any design for this @bigint ?

bigint:
@b4s36t4 You can follow twitter like approach now! aka DMs are more general to all socials!

b4s36t4:
Ok, but you said amount is paid on feature/milestone based. But without assigning how do you manage milestones?

bigint:
Y'all we switched from xmtp to push protocol

XMTP is very limited and slow for huge wallets!

https://github.com/heyxyz/hey/assets/69431456/211639fd-37cb-40e0-8da8-da1ab73930f6


cc @neo773 @b4s36t4

b4s36t4:
Seems push dev network is not working :(

bigint:
@b4s36t4 what is not working, what kind of error you are seeing? can you DM me on discord/telegram 🙇🏼 ?

bigint:
Here are socials https://pastes.io/b6jcjhxjlz

b4s36t4:
@bigint  In favour of moving from xmtp to push I created new PR (#4251)

hashtagobi1:
Hi @bigint,

I hope you are well.

I'd love to give this a go. I've created a brief demo over at:  **[https://chat-firebase-dun.vercel.app/](https://chat-firebase-dun.vercel.app/)**

- Open and Sign in with 2 browsers 
- Test out chat functionality

Stack Used:
- Typescript
- Firebase
- ClerkJS
- Next.JS
- Framer Motion
- Tailwind

/claim #4240

bigint:
@hashtagobi1 we need to use push sdk here 🙇🏼

hashtagobi1:
[...]

PR #4456, opened 2023-12-31
feat: Integrate realtime DMs with Push Protocol [Issue: 4240]

hashtagobi1:
## What does this PR do?

Allows 2 users to communicate via Lens Protocol technologies.

Fixes # (4240)

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Explanation of the changes

Right now I have implemented a basic UI that sends messages. Next step is to integrate communication between lens profiles.

Are Lens profiles are set up already on Hey.xyz? I clicked 'Follow' on a user but when I tried to "Sign in with Lens", the transaction was signed and the page just reloaded however the user was not followed.

Additionally, when I tried to retrieve a lens profile both by 'id' and 'handle' of the users on Hey.xyz, the function returned null.

**My question is this**: how would I retrieve a users Lens profile details from the profile details we have? (e.g. the profile prop in <UserProfile />)

If Lens Profiles don't exist, I can set that up too.
 
===
/claim #4240


apps/web/package.json
```diff
@@ -18,6 +18,7 @@
     "@apollo/client": "^3.8.8",
     "@aws-sdk/client-s3": "^3.484.0",
     "@aws-sdk/lib-storage": "^3.484.0",
+    "@chatscope/chat-ui-kit-styles": "^1.4.0",
```
bigint, line 21: no to third-party designs!

Issue #4182, opened 2023-12-11
Remove hard `location.reload` or hard navigate everywhere


bigint:
Make sure we clear all local cache while logging out!

This excludes `IndexStatus.tsx`

neo773:
Can I get assigned?

bigint:
@neo773 update the description 🙇🏼

kesh-007:
Is the issue still open ?

adevinwild:
Hi @bigint 👋

I'd like to try and tackle this task, it will be my first time on Hey project but I need more information so let's recap very quickly.

If I've understood the problem correctly, when I log out **the page has to reload in order to reset the cache.** (atm)

What you'd like is, instead of having a page that reloads, invalidate the client-side cache without having to reload the page, thus updating all the UI components.

For example, I'm on a post page and I've put a like with my profile
When I log out, the like should disappear BECAUSE I'm no longer logged in, so I'm back to the initial state.

Is this what you are looking for ?

bigint:
exactly @adevinwild

bigint:
@adevinwild you can create test account here https://testnet.tape.xyz/

jayprakash25:
Can I work on this?

adevinwild:
> Can I work on this?

I canceled my attempt on Algora 👍
Good luck, on my side it was impossible to get any replies on Discord from bigint/yoginth, I had one last point before finishing, you can take inspiration from what I did in the PR I think 🙌

PR #4568, opened 2024-01-28
removed hard location.reload everywhere except IndexStatus.tsx , whil…

Deb-cybe:
/claim #4182


bigint:
Closing this, as this does the same, we don't want to take to home page!

Deb-cybe:
> Closing this, as this does the same, we don't want to take to home page!

My bad. Can I attempt again?

PR #4571, opened 2024-01-28
Morefixes

Deb-cybe:
/claim #4182

`This way, each time after getting logged out, a new parameter is added to the URL. In this way it tricks the browser into treating the resource as a new one, prompting it to fetch the updated content from the server.`


Deb-cybe:
@bigint may I know why you closed the issue for this pull request?

PR #4582, opened 2024-01-31
Replace hard location.reload() with alternative.

MohakBajaj:
/claim #4182


bigint:
This is not the solution!

PR #4779, opened 2024-03-26
feat: better tipping data display

### LINKED ISSUE #4778: Tips are counted as Collects

### Current Behavior

When viewing a publication that has received tips the number of tip actions is shows as "collects".

Note how [this post](https://hey.xyz/posts/0xd8-0xf4) is not collectible, but shows 33 "collects". These are actually tips. This is also true for the `/collectors` page.

### Expected Behavior

Hey should distinguish between collect open actions and others, like tipping, and show the proper collect count.

### Steps To Reproduce

1. Create a post with tips enabled, but without making it collectible. 
2. Tip that post. 
3. Observe the erroneous collect count.

Alternatively, view this post: https://hey.xyz/posts/0xd8-0xf4

Web, Mobile

Chrome, Firefox, Brave, Safari, Edge

### Anything else?

_No response_

iPaulPro:
## What does this PR do?

This pull request focuses on better display of tipping and collect stats.

## Related issues

Fixes #4778 

## Type of change

## Explanation of the changes

1. Filters open action counts to only reflect collects.
2. Changes "Action" verb back to "Collect" where appropriate.
3. Adds query to get action count and `actedOn` operation for tips
4. Queries for tip action data for each publication
5. Displays tip counts
6. Shows solid tip icon when acted on
7. Adds `/tippers` sub page for publications

Screen Shot 2024-03-26 at 1 06 18 AM

localhost_4783_posts_0xed-0x19 (2)

localhost_4783_posts_0xed-0x19 (1)

## Side effects

1. The `seedAllowedTokens` script was updated to use Mumbai addresses where available when running on `testnet`.
2. A "solid" variant of the `TipIcon` was created called `TipIconSolid`.
3. A new `ProfileLinkSource` was added when showing `/tippers` called `tips`.
4. A new `UnknownOpenActionData` query was added to get the action count and operation status of the current user for unknown action modules (like tipping).


apps/web/src/components/Publication/OpenAction/UnknownModule/Tip/index.tsx
```diff
@@ -28,6 +32,25 @@ const TipOpenAction: FC<TipOpenActionProps> = ({
...
+
+  useEffect(() => {
+    setHasTipped(
+      tipData?.publication?.__typename === 'Post' &&
+        Boolean(tipData.publication.operations.actedOn.length)
+    );
+    if (tipData?.publication?.__typename === 'Post') {
+      setCountTips(tipData.publication.stats.countOpenActions);
+    }
+  }, [tipData]);
```
bigint, line 53: We can move this on `onCompleted` callback under `useUnknownOpenActionDataQuery`

bigint: Also do we need state for this? can we directly render from the `tipData`? no?

packages/icons/src/TipIcon.tsx
```diff
@@ -14,10 +14,10 @@ export const TipIcon: FC<TipIconProps> = ({ className }) => (
     xmlns="http://www.w3.org/2000/svg"
   >
     <path
-      clip-rule="evenodd"
+      clipRule="evenodd"
```
bigint, line 17: Kudos 👏🏼

bigint (changes requested):

apps/web/src/components/Publication/OpenAction/UnknownModule/Tip/index.tsx
iPaulPro: > Also do we need state for this? can we directly render from the `tipData`? no?

Good point. I'm still a little new to React and hooks and forgot that there's no race condition since `tipData` will actually trigger the query. 🙂 Changes submitted in 29bd91d.

apps/web/src/components/Publication/OpenAction/UnknownModule/Tip/index.tsx
```diff
@@ -37,13 +48,27 @@ const TipOpenAction: FC<TipOpenActionProps> = ({
     return null;
   }
 
+  const hasTipped =
+    tipData?.publication?.__typename === 'Post' &&
+    Boolean(tipData.publication.operations.actedOn.length);
+
+  const countTips =
+    (tipData?.publication?.__typename === 'Post' &&
```
bigint, line 56: Wondering why only post? why not comments, mirrors and quotes?

iPaulPro: In the case of a mirror we're looking at the target publication ID, so it's still be true here if the original post has tips. I don't think `Mirror` type has `operations`, actually. I forgot tipping was enabled for comments and quotes on Hey, though, so I've added those in 2a4305a

apps/web/src/components/Publication/OpenAction/UnknownModule/Tip/index.tsx
```diff
@@ -29,6 +33,13 @@ const TipOpenAction: FC<TipOpenActionProps> = ({
     ? publication?.mirrorOn
     : publication;
 
+  const { data: tipData } = useUnknownOpenActionDataQuery({
+    variables: {
+      module: VerifiedOpenActionModules.Tip,
+      pubId: targetPublication.id
+    }
+  });
```
bigint, line 41: Ok, this is interesting, just found out that it gonna rate limit users. When someone opens a feed and keeps scrolling, it gonna makes 1 extra API call per post. This need to be fixed by somehow using GraphQL alias

packages/lens/documents/queries/UnknownOpenActionData.graphql
```diff
@@ -0,0 +1,40 @@
...
+      operations {
+        actedOn(request: { filter: { address: $module } }) {
+          ... on UnknownOpenActionResult {
+            address
+          }
+        }
+      }
+    }
+  }
+}
```
bigint, line 40: We should not make an extra graphql request for this, instead try using alias inside

`PublicationOperationFields.graphql` and `PublicationStatsFields.graphql`

iPaulPro: So the issue is that aliases aren't fully supported by graphql-codegen. See [this issue](https://github.com/dotansimha/graphql-code-generator/issues/1787). The query type is properly generated, but not the return type. In our case, `PublicationStats` doesn't have the aliased field (even though it's properly returned by the query).

[graphql-typescript-response-factories](https://github.com/homebound-team/graphql-typescript-response-factories) adds support for aliases but only at the top level, not in fragments (see [this issue](https://github.com/homebound-team/graphql-typescript-factories/issues/54)).

Any ideas on the best way to go about this?

apps/web/src/components/Publication/OpenAction/UnknownModule/Tip/index.tsx
iPaulPro: Going to switch this PR to a draft since the current approach is not feasible due to rate limiting and partial alias support in graphql-codegen. ([more here](https://github.com/heyxyz/hey/pull/4779#discussion_r1541760139))

packages/lens/documents/queries/UnknownOpenActionData.graphql
iPaulPro: Thoughts:

- An alias isn't actually needed in `PublicationOperationFields` since `actedOn` is not currently used -- I can move this directly from `UnknownOpenActionData` to `PublicationOperationFields`.
- The publication details page could afford to make the separate query for the count, and keep the "Tippers" page.
- We could do some server-side caching of the tip counts, as a last resort.

If we're unable to show the tip count in the feed, the other parts of this PR are probably still worthwhile. If we cannot find a way to make codegen work with aliases in fragments, or a suitable workaround, caching the tip counts might be a viable option, though not particularly desirable.

iPaulPro:
@bigint I see that tipping was entirely removed, is this PR no longer worth working on?

PR #4812, opened 2024-04-12
feat: switches to Smart Polls

iPaulPro:
## What does this PR do?

Switches to using [EASPollActionModule](https://github.com/iPaulPro/smart-polls) for polls.

## Type of change


iPaulPro:
This is very out of date because it was started well-over a month ago but I'm still waiting on the Open Action to be verified.

One thing missing from this PR, is to use optimistic status for the votes. I'm happy to add that, but didn't want to keep working on it if the action will not be verified in its current state.

bigint:
Hey I think we can move RPC calls to FE rather than using API

bigint:
Also I just invited to the repo so you can create branch here and contribute, I see lot of conflicts already 😅

iPaulPro:
Just a couple updates: EAS isn't yet deployed on Amoy, and I'm still waiting on the poll module to be verified before I continue more work on this PR.

iPaulPro:
This seems like a waste of time since we don't know if EAS will ever launch on Lens Network.

PR #4990, opened 2024-06-17
ER-47: lensPg replacement - publication (apps/api/src/routes/lens/internal/stats/publication.ts)

ecssen-eh:
Updates the /lens/internal/stats/publication endpoint to no longer be reliant on the lensPg database


ecssen-eh:
pls ignore

PR #5236, opened 2024-08-16
fix window install

xiaoyang886:
error "/dev/null cant find" happened when i execute pnpm install


bigint:
@xiaoyang886 we can remove `dev/null`

xiaoyang886:
> @xiaoyang886 we can remove `dev/null`

sure i will change this pr

PR #5237, opened 2024-08-17
fix: remove /dev/null on prisma codegen

xiaoyang886:


bigint (approved):
**LGTM** :shipit: 

LGTM

PR #5240, opened 2024-08-17
fix: include transaction hash in tx frame requests

moesalih:


bigint (approved):
**LGTM** :shipit: 

LGTM

PR #5242, opened 2024-08-18
fix: frame image aspect ratio

moesalih:


bigint (approved):
Epic one @moesalih, thanks for fixing 🙇🏼

PR #5423, opened 2024-10-30
Fix search input reset on enter

Panky55:
Related to #5413


bigint:
@Panky55 will it got reset on enter?

The intention is not to reset on enter 🙇🏼

PR #5564, opened 2025-04-07
Fixes query for collected posts in AccountFeed

iPaulPro:
Collects tab shouldn't filter by author.


bigint:
Great catch 👏🏼 

Was in rush during migration, thanks 🙏🏼

PR #5595, opened 2025-05-12
Fix missing currentAccount check in settings page

Alexhandru:
Logging out from the path "https://hey.xyz/settings" will result in the following:

Screenshot 2025-05-12 131103


bigint (approved):
Thanks a lot for ths fix 💪🏼 

LGTM

PR #6000, opened 2025-07-16
 Remove redundant type annotation in regexLookbehindAvailable IIFE

reject-i:
Removes the redundant boolean type annotation in the immediately invoked function expression (IIFE) of regexLookbehindAvailable.ts. The type is already specified for the variable itself, making the function return type declaration unnecessary. This change improves code readability while maintaining the same functionality and type safety


bigint:
**LGTM** :shipit: 

LGTM

